PR #1203, opened 2013-12-13
Display visual column width in selection

jscipione:
- displays tab width instead of counting hard tabs as 1 character wide,
  which is how TextMate 1 worked.

For example if your tab width is set to 4 and you're on the second
indent position the column will display as 9 (4 \* 2) + 1, before this patch
it would display as 3 (1 \* 2) + 1.


jscipione:
After all these years yes this is still a bug. The main complaint on my patch was that it would take too long to compute on long lines so I've arbitrarily limited this calculation to the first 4096 characters of a line.

PR #1283, opened 2014-12-25
Open favorites at startup and re-activation

estum:
Added a checkbox in the files preference pane which enables showing the
favorites chooser at startup and re-activation with no open documents.
This option is mutually exclusive with «Create one at startup» and
«Create one at re-activated» settings.

screenshot 2014-12-25 23 36 17

### DIFF
--- a/Applications/TextMate/src/AppController Documents.mm
+++ b/Applications/TextMate/src/AppController Documents.mm
@@ -69,8 +69,11 @@ OAK_DEBUG_VAR(AppController_Documents);
 
 - (BOOL)applicationOpenUntitledFile:(NSApplication*)theApplication
 {
-	D(DBF_AppController_Documents, bug("\n"););
-	[self newDocument:self];
+	D(DBF_AppController_Documents, bug("\n"););	
+	if([[NSUserDefaults standardUserDefaults] boolForKey:kUserDefaultsEnableFavoritesAtActivationKey])
+		[self openFavorites:self];
+	else
+		[self newDocument:self];
 	return YES;
 }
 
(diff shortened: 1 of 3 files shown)


sorbits:
I think the behavior is as expected, i.e. I get the favorites (only) when opening TextMate (and there are no windows to be restored) or when clicking the application icon.

I have merged the code changes you did as 71a2c2dfa8a7cc066770dd02fc594e5a4cd8bd2b — I made a few formatting changes and also changed the preferences key to `showFavoritesInsteadOfUntitled` (since I think that is a more apt description).

I did not merge the GUI changes because I think the way it worked was confusing. I would however prefer to get the setting exposed in preferences so I’m leaving your pull request open and plan to revisit it later.

I apologize for the slow response time.

sorbits:
What do you think about changing the behavior so that the favorites instead (optionally) open once the last document window is closed?

It seems to me we want to pick from favorites (recent projects) either when launching TextMate and there is no session to restore, or when closing our (last) project, not really on re-activation.

eyelidlessness:
The on-close behavior is usually unexpected for Mac apps. The typical behavior for a prompt like this is:

- When opening app with nothing to restore
- When nothing is open, clicking the Dock icon
- (Less known, but has the same behavior as clicking the Dock icon) When nothing is open, highlighting the app in command-tab, then holding option before releasing command

I've only ever seen behavior upon closing the last window in apps ported from Windows.

Uvaldoxx:
Delete my email you have the wrong person you dumb mother fucker

sorbits:
@Uvaldoxx If somebody has signed up for a GitHub account with your email address and then subscribed to notifications then you need to contact GitHub staff to get the issue resolved. They are unlikely to see the comments you post here.

You can also perform a password reset for the account then login and disable notifications / change email / delete account.

The account seems to be unused, so probably just do the latter.

Disclaimer: I am not with GitHub staff / support, and I do not control which emails are sent to you from GitHub.

PR #1285, opened 2015-01-30
Fix issue where tabs' rects would overlap in the tab bar on Yosemite.

rdwampler:
I observed this when closing multiple tabs to the right of the active tab on Yosemite.

screen shot 2015-01-30 at 5 48 27 pm

Setting `leftPadding` to 0 fixes the issue (6ea99b4). From studying the code, I am not sure why it was initially set to -1, but of course I might be missing something ;). The other three commits are just some minor changes, which I think make the code a little easier to follow, but feel free to discard them.


sorbits:
The reason for the negative padding is so that the borders collapse.

I have fixed the issue in f17ac29bf1b5a25c13019bec554a01ab23181df9 and merged your other commits, thanks!

rdwampler:
> The reason for the negative padding is so that the borders collapse.

I didn't notice that until I hid the file browser. Thanks!

PR #1286, opened 2015-02-01
fix t_git failed tests

windwiny:
add `sleep` after io::exec("/bin/sh", "-c", script.c_str(), nullptr) != NULL_STR
I don't know why, some time test will raise error, maybe the file not sync to disk, add sleep will pass tests.

    test_scm: 4 of 32 tests failed:
    Frameworks/scm/tests/t_git.cc:105: Expected (wc.status("folder") == scm::status::mixed), found (clean != mixed)
    Frameworks/scm/tests/t_git.cc:135: Expected (wc.status("folder") == scm::status::mixed), found (clean != mixed)
    Frameworks/scm/tests/t_git.cc:150: Expected (wc.status("folder") == scm::status::deleted), found (clean != deleted)
    Frameworks/scm/tests/t_git.cc:178: Expected (wc.status("folder") == scm::status::deleted), found (clean != deleted)

### DIFF
--- a/Frameworks/scm/tests/t_git.cc
+++ b/Frameworks/scm/tests/t_git.cc
@@ -26,6 +26,7 @@ struct setup_t
 		{
 			OAK_FAIL("error in setup: " + script);
 		}
+		sleep(1);
 	}
 
 	scm::status::type status (std::string const& path) const


sorbits:
Thanks for the pull request though calling `sleep` is not solving the actual problem, which is likely a race condition of some sort, it only makes it less likely to occur.

I don’t yet know why exactly there is a race condition with the current code, but I am aware that one exists — it’s only a problem in the tests though.

PR #1287, opened 2015-02-07
fixup! Use `_externalScopeAttributes` as a fallback to export TM_SCM_NAM...

rdwampler:
...E

Added a const reference to avoid the unnecessary copies when using `auto` in range-based for loops. One can find other instances in the code base where we could probably avoid making copies in similar situations, but I did not want to create the patch noise. The intention here is just to correct the obvious mistake and avoid repeating it in the future.


sorbits:
Merged as 438f110dc9d4c551a344729ebbe084188658de8b, thanks!

Since it’s not fixing a bug per se, and the commit it is amending has already been made public, I normally wouldn’t use a `fixup` for this.

PR #1288, opened 2015-02-27
Patch/yosemite look polish

nanoant:
Picture is worth thousand words:

### Before
yosemite-look-polish-before

### After
yosemite-look-polish-after


sorbits:
Merged as b51f7f4...f2d8598, thanks!

I didn’t like putting OS version checks in the project window layout controller as we’re effectively duplicating the tab view styling decisions in this class, so instead I introduced a `neverHideLeftBorder` property in 0e309b1 which we set in the project window layout controller.

PR #1289, opened 2015-03-22
Add "Middle click to close" for tabs

koenpunt:
As a solution for textmate/requests#3

I'm fairly new to Objective-C programming, so I submitted this PR with two different ways (2 commits) of implementing this, because I don't know whats preferred or recommended.

Any feedback is appreciated.

### DIFF
--- a/Frameworks/OakAppKit/src/OakAppKit.mm
+++ b/Frameworks/OakAppKit/src/OakAppKit.mm
@@ -19,7 +19,7 @@ void OakRunIOAlertPanel (char const* format, ...)
 
 BOOL OakIsAlternateKeyOrMouseEvent (NSUInteger flags, NSEvent* anEvent)
 {
-	return ([anEvent type] == NSLeftMouseUp || [anEvent type] == NSKeyDown) && (([anEvent modifierFlags] & flags) == flags);
+	return ([anEvent type] == NSLeftMouseUp || [anEvent type] == NSOtherMouseUp || [anEvent type] == NSKeyDown) && (([anEvent modifierFlags] & flags) == flags);
 }
 
 @interface OakSheetCallbackDelegate : NSObject
--- a/Frameworks/OakAppKit/src/OakTabBarView.mm
+++ b/Frameworks/OakAppKit/src/OakTabBarView.mm
@@ -556,6 +556,11 @@ static NSString* const OakTabItemPasteboardType = @"OakTabItemPasteboardType";
 	[self trySelectTabForView:[self hitTest:mouseCurrentPos]];
 }
 
+- (void)otherMouseUp:(NSEvent*)anEvent
+{
+	[self _performCloseTab:[self hitTest:[[self superview] convertPoint:[anEvent locationInWindow] fromView:nil]]];
+}
+
 - (void)trySelectTabForView:(NSView*)aView
 {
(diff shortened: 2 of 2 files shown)


sorbits:
Thanks, the patch looks good to me, however, the reason we do the “did 
move” check in the regular mouse down function is because the user may 
want to drag a tab, so we do not want to mistake a drag’n’drop for a 
mouse click.

Since dragging is not supported with the “other” button (I assume), 
I think it would be fine to remove this check (and thus the overload of 
otherMouseDown: where we initialize the mouseDownPos variable).

koenpunt:
So 42f698fe28a4de5990ef7d62937157825b6c4e8c would be sufficient? Or should it use the `_performCloseTab` method?

koenpunt:
Also; does this [Frameworks/OakAppKit/src/OakTabBarView.mm#L785](https://github.com/textmate/textmate/blob/master/Frameworks/OakAppKit/src/OakTabBarView.mm#L785) mean there is a modifier key that can be pressed so that closing closes the other tabs instead of itself?

sorbits:
Correct, holding option (⌥) changes it to “close other”.

So it should go through _performCloseTab: to get that behavior (and 
avoid code redundancy).

The method is private, but we’re calling it from our own class, so 
there is no encapsulation violation, if that was your concern.

sorbits:
> So it should go through _performCloseTab: to get that behavior

Actually, the `OakIsAlternateKeyOrMouseEvent` function checks for `NSLeftMouseUp`, we should extend the check to `NSOtherMouseUp` if we do want the ability to option-middle-click a tab to close other (which I assume we want).

Function in question: https://github.com/textmate/textmate/blob/master/Frameworks/OakAppKit/src/OakAppKit.mm#L22

koenpunt:
Add `NSOtherMouseUp` to `OakIsAlternateKeyOrMouseEvent` and merged my previous commits

koenpunt:
About that private method; I just wasn't sure what the general event flow is. Because I've seen dispatching in a multitude of ways, including the (global?) delegate, NSApp and IBAction's.

sorbits:
Merged as 8829637...808aaab, thanks!

I tweaked the commit message a bit and also added a check for `NSOtherMouseUp ` that should delay resizing the tabs to the left of the one being closed when closing with the middle mouse button (untested, as I don’t have an input device with a third button).

As for dispatching; seems 6748c281098d87ac3c441efe0de7a9d1d38facdc made it inconsistent, having `_performCloseTab:` send directly to the delegate, with `performClose:` still using `NSApp`.

I’ve updated `performClose:` to also send directly to the delegate in commit 31464f39fdb33ea449d983bf4ffa085d68068b8d, thanks for bringing that to my attention.

sanssecours:
Hi Koen and Allan,

thank you for implementing this useful feature.

> (untested, as I don’t have an input device with a third button).

You can simulate a middle click via [BetterTouchTool](http://www.bettertouchtool.net). Here is my configuration, which maps a four finger tap to a middle click:
bettertouchtool

Kind regards,
  René

sorbits:
Thanks for the tip, I was able to test the patch with this tool :)

koenpunt:
I actually used MagicPrefs to test, but with same results :)

PR #1291, opened 2015-03-27
Add missing shadow to find/replace history pop-up list

yuuq:
The NSComboBox, whose pop-up list has shadow, and so it will be nice for this similar control.

before:
without shadow

after:
with shadow


sorbits:
Thanks, I made the change in the xib (Interface Builder) rather than in the code: 5a1754bc0198d9cc03899265127a6cf19f9c5a5e

PR #1292, opened 2015-03-31
moving a comma that seems misplaced

waynn:


sorbits:
Merged as 9d51774, thanks!

PR #1294, opened 2015-04-08
Use googles dense-hash-map for theme cache speedup

joachimm:
Googles sparse-hash-map library contains a dense-hash-map. It is about 60-70% faster than c++ std::map, when looking up Textmate's scope::scope_t.

This patch switches the theme scope cache from std::map to google::dense_hash_map.


sorbits:
Merged as 50250bb...70893f4, thanks!

PR #1295, opened 2015-04-12
Fixed fatal error in Applications/indent/indent.cc

fish2000:
The static `std::map` instance (named just `map`, on line 79) had its initializer template params in the wrong order, which kept it from compiling. Swapping them as per this PR lets `indent` build and run with no issues.


sorbits:
Merged as 2350938, thanks!

PR #1296, opened 2015-04-23
Fix compile error and a couple minor warnings with newer version of clang

rdwampler:


sorbits:
Merged as 2350938...d712167, thanks!

PR #1299, opened 2015-08-07
Use pthread method for thread local storage

rdwampler:
This effectively undoes 172ce9d4282e408fe60b699c432390b9f6e3f74a. It seems Apple discourage static linking and the new version of clang in Xcode 7 issue warnings when the target builds are different.


sorbits:
Merged as 60050cf...14c6a70, thanks!

PR #1300, opened 2015-08-07
Silence a couple of false warnings with about unused local typedefs

rdwampler:
The new release of clang with Xcode 7 now turns on warnings for unused local typedefs. This triggers warnings in OakTextView and the sparsehash library, but they seem to be false positive. This PR silence them nonetheless, since in the case of the sparsehash library the output is very noisy.


sorbits:
Merged as 14c6a70...2c6ad9b, thanks!

PR #1301, opened 2015-09-19
Fix some build errors with the new 10.11 SDK

rdwampler:
This resolves most of the errors when building with the new SDK. The only remaining issue is that the headers for `openssl` are no longer included. For now, I just installed the missing headers via brew and copy them over when building. I don't know if you want go with the above approach or update the code to use `CommonCrypto`/`Security Transforms`.

As for 9e4e88c, I submitted a radar but I have not heard back.

### DIFF
--- a/Applications/TextMate/src/AboutWindowController.mm
+++ b/Applications/TextMate/src/AboutWindowController.mm
@@ -242,7 +242,11 @@ static NSTextField* OakCreateTextField ()
 
 // ============================
 
+#if !defined(MAC_OS_X_VERSION_10_11) || (MAC_OS_X_VERSION_MAX_ALLOWED < MAC_OS_X_VERSION_10_11)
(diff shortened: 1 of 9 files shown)


rdwampler:
I removed the update to dialog submodule in the first commit and clarify the "new protocols" commit.

sorbits:
Thanks!

For 3f5575c I think it was supposed to have used the `filePathURL` method, but just accessing `path` directly should be fine (as you write),

As for openssl, the problem is that I use some lower level functions for the license key validation that I don’t think is available in any of the crypto frameworks that Apple provide.

rdwampler:
You're correct it doesn't appear to be any replacement for openssl using Apple's frameworks. The documentation suggests to provide your own version of openssl and statically link against it, but I am not having much getting it to statically link though.

PR #1302, opened 2015-09-25
Fix typo in 2015-09-16 (v2.0-beta.8) section of release notes

ram-nadella:


sorbits:
Merged as acd4497, thanks!

PR #1303, opened 2015-10-01
[RFC] Replace system provided openssl with LibreSSL

rdwampler:
This is follow up to #1301. 

I thought about having the build default to use the system headers if present (e.g., building on 10.10 or older), but I think it easier instead to just require a more recent version of the openssl headers and statically link it.

I have not tested this with Macports so I don't know if the paths in `configure` are right (I plan on testing it in a few days). Thoughts?

### DIFF
--- a/Frameworks/license/src/license.cc
+++ b/Frameworks/license/src/license.cc
@@ -5,12 +5,6 @@
 #include <text/tokenize.h>
 #include <text/case.h>
 #include <text/format.h>
-
-#ifdef DEPRECATED_IN_MAC_OS_X_VERSION_10_7_AND_LATER
-#undef DEPRECATED_IN_MAC_OS_X_VERSION_10_7_AND_LATER
-#define DEPRECATED_IN_MAC_OS_X_VERSION_10_7_AND_LATER
-#endif
-
 #include <openssl/bio.h>
 #include <openssl/pem.h>
 #include <openssl/rsa.h>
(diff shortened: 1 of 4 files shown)


sorbits:
I think this is an excellent workaround. Although we do not use `openssl` for security critical stuff, using the latest version is preferred over Apple’s abandoned version, and running `brew install libressl` is hardly an issue for people who want to build TextMate.

One issue I noticed is that clang warns that `libressl` is built for a newer version of the OS than TextMate itself. I assume that in theory, `libressl` could use OS functionality not available on 10.7, but it’s probably highly unlikely.

Unfortunately I recently deleted my 10.7 boot image, but I’ll try to get it tested there, just to be sure.

rdwampler:
> One issue I noticed is that clang warns that libressl is built for a newer version of the OS than TextMate itself.

I noticed this too for `capnp`. If by chance these are using newer OS functionality, we could provide a section in the README, to show how to edit the brew formula.

For `capnp`, one would need to change the line from:
             
       system "cmake", "..", *std_cmake_args

to:

      system "cmake", "-DCMAKE_OSX_DEPLOYMENT_TARGET=10.7", "..", *std_cmake_args

 and then `brew install --build-from-source capnp`.

 For LibreSSL, the change is larger and requires building again the current HEAD.

sorbits:
Merged as acd4497...a23b522, thanks!

The deployment target is only relevant when distributing the resulting binary (to someone using an older OS), i.e. only I should really need to care about this.

PR #1304, opened 2015-10-02
Refresh when unfolding on mouse down event

rdwampler:
This seems to be necessary when the document occupies less space than the visible document window.


sorbits:
Merged as 0547479, thanks!

PR #1307, opened 2015-10-25
Add "Commit and Continue" button to the commit window

jacob-carlborg:
The use case for this feature is when an interactive rebase is in
progress and be able to, in one command, amend the current commit
and continue with the rebase.

By default, if the --rebase-in-progress flag is passed when the
"commit" command is invoked the commit button will be in a
"commit and continue" state. This is indicated by the title of the
commit button now being "Commit & Continue". If the option (⌥) key
is being hold down the state will revert back to the standard "commit"
state and the title will change to "Continue".

If the butten was clicked when it was in the "commit and continue"
state the "commit" command will print "TM_SCM_COMMIT_CONTINUE=1"
(without the quotes) to stdout as the last line. If the button was
clicked when it was in the standard "commit" state the output may
contain "TM_SCM_COMMIT_CONTINUE=0", or the "TM_SCM_COMMIT_CONTINUE"
string will be absent

### DIFF
--- a/Applications/commit/src/commit.mm
+++ b/Applications/commit/src/commit.mm
@@ -33,7 +33,12 @@ static double const AppVersion = 1.0;
 		fprintf(stderr, "%s", [err UTF8String]);
 
 	if(NSString* out = someOptions[kOakCommitWindowStandardOutput])
+	{
 		fprintf(stdout, "%s", [out UTF8String]);
+
+		if([someOptions[kOakCommitWindowContinue] boolValue])
+			fprintf(stdout, "TM_SCM_COMMIT_CONTINUE=1\n");
+	}
 
 	_returnCode = [someOptions[kOakCommitWindowReturnCode] intValue];
 
--- a/Frameworks/CommitWindow/src/CommitWindow.h
+++ b/Frameworks/CommitWindow/src/CommitWindow.h
@@ -7,6 +7,7 @@ static NSString* const kOakCommitWindowEnvironment          = @"environment";
 static NSString* const kOakCommitWindowStandardOutput       = @"stdout";
 static NSString* const kOakCommitWindowStandardError        = @"stderr";
 static NSString* const kOakCommitWindowReturnCode           = @"returnCode";
+static NSString* const kOakCommitWindowContinue             = @"continue";
 
 @protocol OakCommitWindowClientProtocol <NSObject>
 - (void)connectFromServerWithOptions:(NSDictionary*)someOptions;
--- a/Frameworks/CommitWindow/src/CommitWindow.mm
+++ b/Frameworks/CommitWindow/src/CommitWindow.mm
@@ -117,6 +117,7 @@ static void* kOakCommitWindowIncludeItemBinding = &kOakCommitWindowIncludeItemBi
 @property (nonatomic) OakCommitWindow*                   retainedSelf;
 
 @property (nonatomic, readonly) BOOL                     rebaseInProgress;
+@property (nonatomic, readonly) BOOL                     amend;
 @end
 
 @implementation OakCommitWindow
@@ -146,7 +147,7 @@ static void* kOakCommitWindowIncludeItemBinding = &kOakCommitWindowIncludeItemBi
 		self.window.releasedWhenClosed = NO;
 		self.window.frameAutosaveName  = @"Commit Window";
 
-		_commitButton = OakCreateButton(@"Commit", NSRoundedBezelStyle);
+		_commitButton = OakCreateButton([self commitTitle], NSRoundedBezelStyle);
 		_commitButton.action                    = @selector(performCommit:);
 		_commitButton.keyEquivalent             = @"\r";
 		_commitButton.keyEquivalentModifierMask = NSCommandKeyMask;
@@ -229,6 +230,11 @@ static void* kOakCommitWindowIncludeItemBinding = &kOakCommitWindowIncludeItemBi
 
 	if(self.rebaseInProgress)
 		[NSEvent removeMonitor:_eventMonitor];
+}
+
+- (NSString*)commitTitle
+{
+	return self.amend ? @"Amend" : @"Commit";
 }
 
 - (NSDictionary*)allViews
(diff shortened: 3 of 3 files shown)


jacob-carlborg:
Updated, removed the environment variable and added the `--show-commit-and-continue-button` flag instead.

I would like to have a key equivalent for the button but could come up with a good one. Any suggestions?

sorbits:
How about instead of adding an extra button, we instead change Commit 
→ Commit & Continue when a rebase is in progress?

We could support holding option (⌥) to revert the button back to only 
doing a commit, but I’m not sure there is ever a reason not to call 
“git rebase --continue” after a commit, when a rebase is in 
progress. I.e. the only other action that makes sense in this state is 
“git rebase --abort”, right? So presumably the only reason for going 
with a simple commit is to get the “git rebase --continue” output in 
a terminal rather than TextMate’s HTML output view.

Btw: I notice that all your if’s have space before the opening 
parenthesis.

koenpunt:
> I.e. the only other action that makes sense in this state is
“git rebase --abort”, right?

What about partial commits? I often use rebase to edit and split a commit.

jacob-carlborg:
Yeah, and it's possible to amend and commit multiple times before continuing.

jacob-carlborg:
> How about instead of adding an extra button, we instead change Commit → Commit & Continue when a rebase is in progress? We could support holding option (⌥) to revert the button back to only doing a commit

I like that idea. I started to implementing this but I can't figure out how to allow both ⌘Enter and  ⌘⌥Enter as key equivalents for the commit button.

rdwampler:
Have a look at `FFResultViewController` (https://github.com/textmate/textmate/blob/master/Frameworks/Find/src/FFResultsViewController.mm#L258). You should be able to reset the key equivalent for the commit button when your handler is called.

jacob-carlborg:
Update to only use one button. When the `--rebase-in-progress` flag is passed it will default to `Commit & Continue`. The ⌥ key can be used to revert back to the previous, `Commit`, behavior. @rdwampler thanks for the help with that.

I also added a second commit which sets the title of the commit button to `Amend` if the `--amend` flag is passed.

sorbits:
Thanks for the update, using “Amend” is a nice touch :)

I was wondering if it would make sense to try and generalize this? So rather than have an option per possible button title, we would simply tell the window what the titles should be (possibly using `|` to specify an alternate title). Although right now, the _Commit_ button will tel how many files are about to be committed, so perhaps this is not so straight forward.

I was also wondering if perhaps we should take the opportunity with these new options to change the output format from the commit window command. So if none of the options are specified, it would output the old line with `-m «message»` and files, where using any of the new options would switch to instead returning a property list with the commit message, the files, and which button was pressed.

As for the latter, it appears that the commit window will output `TM_SCM_COMMIT_CONTINUE=0` even when none of the new options are used. I’m quite sure this will break most SCM bundles.

sorbits:
My thoughts about changing output format and generalizing the button titles can wait, with the above changes I’ll gladly merge it.

I can also make the changes myself if you agree with them.

martham77:
Thank you this is very interesting and just starting to explore it and want to learn more about it. .

jacob-carlborg:
> My thoughts about changing output format and generalizing the button titles can wait, with the above changes I’ll gladly merge it.

Yeah, I think that can wait. This started out as a simple enhancement. I don't want this to get too big.

@sorbits please comment in the pull request next time and not on the commits. It's easier to keep track of then.

jacob-carlborg:
I agree with the output format. It didn't feel quite right to add `TM_SCM_COMMIT_CONTINUE` at the end of the output.

jacob-carlborg:
> I can also make the changes myself if you agree with them.

No, I can make the changes.

jacob-carlborg:
Update, I think I fixed all comments. I decided to only add the event monitor if the there's a rebase in progress.

sorbits:
Merged as 8d8c539...bdcba4d.

I changed the event monitor to use a weak reference to self (and not itself be weak) as it was otherwise retaining our object (that removed it in dealloc, thus leaking the entire commit window). This is also an issue with the find window result controller (where you copied the code from), but since that is a singleton, it will never be released (although I’ll fix it there as well).

I added two additional commits changing the options to `--show-continue-button` and `--commit-button-title «title»` respectively, let me know if you disagree with this naming (I found `--rebase-in-progress` too descriptive of the use-case rather than what was requested from the window, and `--amend` was just begging to be generalized).

jacob-carlborg:
> I changed the event monitor to use a weak reference to self (and not itself be weak) as it was otherwise retaining our object (that removed it in dealloc, thus leaking the entire commit window). This is also an issue with the find window result controller (where you copied the code from), but since that is a singleton, it will never be released (although I’ll fix it there as well).

Ok, I see. As you said, I just copy pasted the code :smiley:.

> I added two additional commits changing the options to --show-continue-button and --commit-button-title «title» respectively, let me know if you disagree with this naming (I found --rebase-in-progress too descriptive of the use-case rather than what was requested from the window, and --amend was just begging to be generalized).

I think `--commit-button-title` is a bit misleading since it's only the prefix and not the complete title. If you would like to generalize this even more and support the complete title you could allow some form of template syntax in the title:

```
commit --commit-button-title "Commit %{count} Files%{continue_suffix}"
```

I forgot to update the commit message. Now it says that the output may contain `"TM_SCM_COMMIT_CONTINUE=0"`.

Also, I would have cleaned up the commits a bit. Now two new commits were added and then immediately changed in the next two, which you don't seem to like (and I agree) :smiley:.

sorbits:
I think putting the full button title in the argument is problematic 
because we will need syntax for singular and plural, each bundle can end 
up having a full text string, which is redundant, e.g. I changed File(s) 
→ Item(s), which would then have to be done in each bundle, and I 
consider adding “ & Continue” to the title, when we ask for a 
continue button, an “implementation detail”, e.g. we could later 
decide to do this differently.

As for the argument itself, I did consider sonething like 
--commit-action-verb but since we already have --action-cmd as argument, 
I felt it was best to avoid the word “action”.

As for changing the arguments in the following commits, normally I would 
have squashed them, but I felt this was too subjective to do in your 
name, and also, the --ammend change did some refactoring wrt. to the 
button title, so incase I introduced an issue here, it was better to 
have it broken out.

I am generally not overly cautious about not force-pushing to the master 
branch, so if you have things you want to cleanup, do fixup! commits and 
I can do the rebase/force-push :)

PR #1308, opened 2015-12-28
Use panel language unless spellingLanguage is specified

nanoant:
This partially reverts changes of 3fdc72b93af0167e234c48314658d702f65a8a5f: Support spell checking being “automatic by language”.

Before 3fdc72b `spellingLanguage` .tm_properties setting (default "en") was the only way to set spelling language for file (buffer). English was default language for all files. 3fdc72b introduced change that when “automatic by language” was selected in system's spelling panel then TextMate was ignoring `spellingLanguage` setting and was using automatic by language.

However because “automatic by language” is default on OS X, effectively 3fdc72b makes `spellingLanguage` setting no longer effective. To make it work one needs to set explicit spelling language either in System Preferences or in spelling panel upon each TextMate run (since this is getting reset after application restart) - which is counter-intuitive, can be treated as regression and it is vaguely described in ChangeLog:

 * Support spell checking being “automatic by language”. This is set via the
   spelling panel.

This change presents alternative approach, introducing new empty `spellingLanguage` setting "" (which is now default), which makes use system panel language setting, including “automatic by language”.

From now on all files will be checked against system panel selected language (or automatic), unless .tm_properties project specifies explicitly language for given file using, eg.:

    [ locale-en_US.ini ]
    spellingLanguage = en_US

    [ locale-pl_PL.ini ]
    spellingLanguage = pl_PL

Or automatically depending on file name:

    [ locale-*.ini ]
    spellingLanguage = '${TM_FILEPATH/^.*locale-([a-z]+_[A-Z]+).*$/$1/}'

### DIFF
--- a/Applications/TextMate/src/AppController Menus.mm
+++ b/Applications/TextMate/src/AppController Menus.mm
@@ -97,6 +97,17 @@ OAK_DEBUG_VAR(AppController_Menus);
 		ordered.emplace(to_s(str ?: lang), lang);
 	}
 
+	NSString* systemSpellingLanguage;
+	if ([spellChecker automaticallyIdentifiesLanguages])
+		systemSpellingLanguage = @"Automatic by Language";
+	else
+	{
+		systemSpellingLanguage = (NSString*)CFBridgingRelease(CFLocaleCopyDisplayNameForPropertyValue(CFLocaleGetSystem(), kCFLocaleIdentifier, (__bridge CFStringRef)[spellChecker language]));
+	}
+	NSMenuItem* menuItem = [aMenu addItemWithTitle:[NSString stringWithFormat:@"System (%@)", systemSpellingLanguage] action:@selector(takeSpellingLanguageFrom:) keyEquivalent:@""];
+	D(DBF_AppController_Menus, bug("Represented Object: "" (System)\n"););
+	menuItem.representedObject = @"";
+
 	for(auto const& it : ordered)
 	{
 		D(DBF_AppController_Menus, bug("Add Item: %s\n", it.first.c_str()););
--- a/Frameworks/buffer/src/buffer.cc
+++ b/Frameworks/buffer/src/buffer.cc
@@ -11,7 +11,7 @@ OAK_DEBUG_VAR(Buffer_Parsing);
 
 namespace ng
 {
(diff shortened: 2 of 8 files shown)


sorbits:
If the user pick a spelling language via the UI then (with this change) there is no way to get back to “automatic by language” (other than manually editing the `Global.tmProperties` file).

So for this change I think we need to add an “Automatic by Language” option to the Edit → Spelling submenu.

nanoant:
@sorbits Yeah, that makes sense. I'll add a commit for this menu item.

nanoant:
@sorbits Updated according to your suggestion.

sorbits:
Sorry for taking so long with this one.

For completeness I think we need an additional (explicit) “Automatic by Language”, so the user can pick either “system” (which defaults to “Automatic by Language”) or explicitly pick this, since the system setting could be changed (or spelling panel could be changed locally).

Do you agree with this (I’ll add this if you agree)?

It’s all a bit confusing because we have this spelling panel with its own independent setting, would be best to have them in sync, but we lack the ability to be notified about changes in the panel or notification about when the panel is brought to front etc. (unlike e.g. the font panel which will adapt to the current first responder).

nanoant:
> Sorry for taking so long with this one.

No problem. Thanks for all your work on TM2.

> For completeness I think we need an additional (explicit) “Automatic by Language”, so the user can pick either “system” (which defaults to “Automatic by Language”) or explicitly pick this, since the system setting could be changed (or spelling panel could be changed locally).
> Do you agree with this (I’ll add this if you agree)?

Yeah I do agree, but I understand new (explicit) “Automatic by Language” will be be reflected with some new `"auto"` setting, because default empty `""` means system in the PR above?

So altogether the menu and the `language` setting values will look like:

* System (Current system setting) -> `""`
* Automatic by Language -> `"auto"`
* Engligh -> `"en"`
* German -> `"de"`
(...)

Correct?

sorbits:
> Correct?

Correct!

sorbits:
Merged as 9c3923b...272f691.

I didn’t yet do the “automatic by language” menu item because I think we need to introduce some abstraction, as the spell checker ought to have its language synced whenever it is used, for example before we use the “learnWord:” method, though the direction of sync is not obvious, should probably take the spell checker’s language when the panel is visible, and otherwise push TM’s language to the panel, though that does make the “system” setting somewhat redundant…

PR #1309, opened 2015-12-30
Double Click to Select Typing Pairs

briantkelley:
TextMate highlights certain character pairs (e.g. { }) when the caret passes over either of the characters. It can be useful to double click to select the text in between and including the caracter pairs for deleting code, copying code, for visually differentiating the affected code, etc. Update the meaning of kSelectionExtendToWord to consider a character pair, and the text in between, a "word".

This patch is free and released into the public domain.


sorbits:
Thanks for the patch and sorry about the late response.

I noticed a problem with the patch, namely that it tests if `from` is at a typing pair character _after_ we perform the normal word selection (which can modify both `from` and `to`).

In most cases it still works as desired, but try e.g. double-click at the right of a line ending with empty braces (`()`). For me I get the empty braces selected, but the selection continues to the next closing brace in the document.

This can be fixed but we run into another issue, namely that this patch changes the behavior of word selection, take the following 3 examples (where `‸` indicates our caret position):

1. `foo‸(bar, baz)` — <kbd>⌃W</kbd> selects `foo`
2. `foo(‸bar, baz)` — <kbd>⌃W</kbd> selects `bar`
3. `foo(bar, baz‸)` — <kbd>⌃W</kbd> selects `baz`

In example 1 and 3 our caret is on a typing pair character, i.e. this is where the patch should select the full function parameters (`(bar, baz)`), but I do not think we want this behavior when using the keyboard to select words (should we desire to select the function parameters, we can always use <kbd>⇧⌘B</kbd>).

So I think we need a separate function used with double-click that checks if we double-click a typing pair character, rather than change the behavior of word selection, although I do like that in the case that there is whitespace before/after the typing pair character, <kbd>⌃W</kbd> will effectively act as <kbd>⇧⌘B</kbd>.

briantkelley:
No worries about the late response, my apologies for not responding sooner :). I think separating the double click behavior from word selection behavior makes sense, which I figured would be a point of discussion. I hope to find time in the next week or so, and I'll take a look at the brace problem too. Thanks!

briantkelley:
So sorry for taking _forever_ to update the patch, but I think the new implementation is much better.

I fixed the logic when calculating the typing pair range so I believe all the over-selection problems are fixed. I added a new selection action to select either a typing pair or a word that's only used when double clicking in the text view. All the keyboard shortcuts should be unaffected by this change.

Let me know what you think. Now <kbd>^W</kbd> will no longer effectively act like <kbd>⇧⌘B</kbd>, but it shouldn't be too hard to update the typing pair re-selection logic if you think that behavioral change makes sense. Thanks!

sorbits:
Thanks, I made a few changes and merged as 2dd8255.
Please let me know if you find any problems with the revised version.

One change was that you had changed both the mouse-click and mouse-drag action to use the new “word or typing pair”, however, for the mouse-drag, the action didn’t really work as expected, so I made double-click-and-drag use the old word selection method. This does lead to a minor issue if you double-click a typing pair and then (while holding mouse) starts to drag, it then switches to word selection, but I think this behavior is acceptable.

The other change is that we now only look for typing pairs (to the left/right of where we click) if there is no existing selection, and we bypass the code that potentially does a selection, i.e. we start out by looking for typing pairs to the left/right of caret, then does a “typing pair selection” if any is found, and then continue to word selection (if this failed).

PR #1310, opened 2016-01-02
Break on comments extending selection to paragraph

nanoant:
Previously when reformatting paragraph comment lines were selected as paragraph
lines, this have led to mixing comment content into reformatted paragraph
content breaking the syntax of reformatted code, eg.:

~~~bash
some very long ... line of text
# some comment
some other long ... line of text
~~~

As an effect we got:

~~~bash
some very long ...
... text # some comment some other ...
... line of text
~~~

The problem described above was especially visible when using <kbd>⌃Q</kbd> to Reformat
block of Git commit message, when below of the typed text there was a Git
default comment. Also this problem could be noticeable by authors using LaTeX.

This change checks whether a candidate line for paragraph extension has a
"comment.line" scope at the end and breaks upon such trailing line comment.

Doing Select Paragraph <kbd>⌃⌥P</kbd> again will select the comment as well.

@sorbits I know this solution is not perfect and quite hacky, but I couldn't think of anything more generic.

### DIFF
--- a/Frameworks/selection/src/selection.cc
+++ b/Frameworks/selection/src/selection.cc
@@ -582,10 +582,12 @@ namespace ng
 			{
 				if(line == 0)
 					return 0;
-
 				for(size_t n = line-1; n > 0; --n)
 				{
-					std::string const& line = buffer.substr(buffer.begin(n), buffer.end(n));
+					size_t eol = buffer.end(n);
+					if(plist::is_true(bundles::value_for_setting("excludeFromParagraphSelection", buffer.scope(eol-1))))
+						return buffer.begin(n+1);
+					std::string const& line = buffer.substr(buffer.begin(n), eol);
 					if(text::is_blank(line.data(), line.data() + line.size()))
 						return buffer.begin(n+1);
 				}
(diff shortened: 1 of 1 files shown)


nanoant:
@sorbits Any chance to have this reviewed anytime soon? I can see there're new betas, but this PR hasn't been included. I think this change is really useful especially when editing Git messages or working with LaTeX.

sorbits:
I commented on this via IRC but you probably missed it then.

I wasn’t keen on hardcoding this to a scope and would prefer to see it done via scoped preferences similar to how the character class system works.

So instead of using `commentLineScope.does_match(…)` you would do something like `plist::is_true(bundles::value_for_setting("excludeFromParagraphSelection", buffer.scope(eol)))` and then e.g. in the Source bundle we can set `excludeFromParagraphSelection = :true` for the `source.comment` scope.

nanoant:
@sorbits I modified this PR according to your suggestions. Also there is accompanying PR textmate/source.tmbundle#20.

sorbits:
Merged as aba1eb42ace2c17b42974c245779256d6d44eaad, thanks!

I made a minor change in that rather than set `eol = buffer.end(n)` and use `eol-1` we set `eol = buffer.eol(n)`. This removes the assertion that there is always one (newline) character at the end of the line. It does exclude the newline from `text::is_blank(…)` check, but that shouldn’t matter.

PR #1311, opened 2016-01-15
Add prototype for bug macro

mkhl:
Compiling code that uses the `bug` macro
while the `NDEBUG` symbol is defined
(which it is by default)
results in “expression result unused” warnings
for each argument to the `bug` call.

This change adds the parameter declaration
from the debug version to the non-debug version,
which silences these warnings.


sorbits:
Merged as 5960f13, thanks!

PR #1312, opened 2016-01-15
Travis

mkhl:
This pull request includes
* a build configuration for Travis
* a fix for the `gen_credits` problem encountered while attempting to build on Travis
* running tests synchronously by default
* a way to configure default test flags when running `./configure`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: objective-c
+before_install:
+  - brew update
+  # boost and hg are already installed
+  # update boost just to be sure
+  - brew outdated boost || brew upgrade boost
+  - brew install ragel multimarkdown ninja capnp google-sparsehash libressl
+install: ./configure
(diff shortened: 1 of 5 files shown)


sorbits:
Merged as 5960f13...a7d069f, thanks!

sorbits:
I have made parallel test execution the default again (in 7dbb44c4a068ce79b92eebdfd8026202bac257d0) since this does not give me any problems anymore (the involved code has been changed to use `CFRunLoopPerform()` with the run loop of the caller instead of `dispatch_async` using `dispatch_get_main_queue()`, the latter being safe in TextMate, where caller is always executing on the main queue, but not the case for tests, even when they are running synchronously).

sorbits:
Meant to add that it can be disabled with `--no-parallel`, but if it is giving a problem then do let me know.

PR #1313, opened 2016-01-27
Fixes a crash on "force touch"

bSr43:
This PR fixes a crash which occurs when one is doing a "force touch" on a trackpad when the cursor is over an empty area of a plain text document.

It seems that the "showDefinitionForAttributedString:" can accept a nil argument, but not an empty string. The inserted test is preventing the call when the extracted string contains no characters.


perajovic:
Please, merge it as soon as you can. I have also problem with this.

sorbits:
Merged as 041967d (I made the check on the string, instead of the range).

I had noticed the new crashes from this API but had yet to figure out the problem, so thank you very much for this fix!

PR #1314, opened 2016-01-27
Ensure path exists when adding document to Recent menu

rdwampler:
It's possible that the path on the document is set even though it does not exist on disk. E.g., when using `mate` to create a new document from the command line.

I tried to handle this in `mate.cc` by treating such files as temporaries and not tracking them, but then they would not be added to the Recent menu after saving.


sorbits:
Merged as 9c3923b, thanks!

PR #1315, opened 2016-02-02
CommitWindow: Force update of constraints after animation is completed

rdwampler:
This is necessary to prevent resizing of the document view when removing and then immediately adding a new constraint to the document view.

I don't recall seeing this issue before so it may just occur on 10.11.


sorbits:
Merged as 87df096, thanks!

PR #1316, opened 2016-02-02
RFC: Jump to marks

rdwampler:
This essentially add keyboard shortcuts to jump to next/previous mark, regardless if it's a bookmark or not. I do not know if these are the best shortcuts to use. Xcode uses <kbd>⌘</kbd> + <kbd>' </kbd> and  <kbd>⌘</kbd> + <kbd>" </kbd>, but those are already being use by other bundles. And I couldn't find examples from other apps that have a similar feature. Any suggestions?

### DIFF
--- a/Applications/TextMate/resources/English.lproj/MainMenu.xib
+++ b/Applications/TextMate/resources/English.lproj/MainMenu.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="9059" systemVersion="14F1021" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="9531" systemVersion="15E27e" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none">
     <dependencies>
         <deployment version="1070" identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="9059"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="9531"/>
     </dependencies>
     <objects>
         <customObject id="-2" userLabel="File's Owner" customClass="NSApplication">
--- a/Applications/TextMate/resources/English.lproj/MainMenu.xib
+++ b/Applications/TextMate/resources/English.lproj/MainMenu.xib
@@ -836,6 +836,17 @@ CA
                                 <menu key="submenu" title="Go to Bookmark" id="595"/>
                             </menuItem>
                             <menuItem isSeparatorItem="YES" id="439"/>
+                            <menuItem title="Jump To Next Mark" keyEquivalent="&gt;" id="gGW-Hn-MX6">
+                                <connections>
+                                    <action selector="jumpToNextMark:" target="-1" id="999"/>
+                                </connections>
+                            </menuItem>
+                            <menuItem title="Jump To Previous Mark" keyEquivalent="&lt;" id="5Mq-52-EUe">
+                                <connections>
+                                    <action selector="jumpToPreviousMark:" target="-1" id="998"/>
+                                </connections>
+                            </menuItem>
+                            <menuItem isSeparatorItem="YES" id="HcX-yL-1Rc"/>
                             <menuItem title="Scroll" id="429">
                                 <modifierMask key="keyEquivalentModifierMask"/>
                                 <menu key="submenu" title="Scroll" id="430">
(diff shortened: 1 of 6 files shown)


koenpunt:
Marks as in bookmarks, git marks, error marks, etc? Isn't that <kbd>F2</kbd>?

rdwampler:
> Marks as in bookmarks, git marks, error marks, etc? 

I mainly had "issues" in mind (i.e., warnings, errors, and notes) like Xcode's "Jump to Next/Previous Issue". However, I do find it useful to have a way to quickly jump to the git marks when making many small changes through out a file.

> Isn't that F2?

<kbd>F2</kbd> is currently only for bookmarks. I didn't want to overload this shortcut to jump to other marks like the git, error, warning and note marks since bookmarks are used to quickly return to a particular place in the document. If there were a lot of marks (e.g., git marks) then it would defeat the purpose.

perajovic:
It should be very convenient if we can have jump between previous/next cursor position. But don't know if that is doable within this PR.

koenpunt:
<kbd>F2</kbd>  / <kbd>SHIFT</kbd> <kbd>F2</kbd> do forward and backward for bookmarks. What about <kbd>CTRL</kbd> <kbd>F2</kbd> / <kbd>SHIFT</kbd> <kbd>CTRL</kbd> <kbd>F2</kbd> for other marks?

sorbits:
Thanks, merged as 42b6c04...18a79d3 (I reworked the commits a bit so that `nil` as mark type means “all marks” rather than having separate functions for searching all of them).

As for keyboard shortcut, since both <kbd>⌘></kbd> and <kbd>⌘<</kbd> are available, I think this is a fine choice. What I have been wanting to do is add the marks to <kbd>⇧⌘T</kbd> (as a separate source), this would also give some options about what type to search for. I do think including the git marks are a bit excessive (for these, we could treat successive marks as a single mark to cut down on the marks <kbd>⌘></kbd> visits), but I am not sure how to best decide what marks to skip, possibly this could be done by naming convention, as marks can contain a payload, will have to ponder this a bit more…

infininight:
Got a email that this shortcut conflicts with the Move Focus to Next Window shortcut on the Swedish keyboard layout. This is the shortcut that is <kbd>⌘`</kbd> on the US and other layouts. Not sure if there is any better shortcut to choose however, or how many layouts this has a conflict with.

rdwampler:
> Not sure if there is any better shortcut to choose however, or how many layouts this has a conflict with.

We probably should change it since it is a system wide shortcut. Also it looks like it conflicts with quite a few keyboard layouts (@sorbits does this also conflict with the Danish keyboard layout?).  Maybe we can change it to <kbd>⌃,</kbd> and <kbd>⌃.</kbd>, idk.

sorbits:
I am using a U.S. keyboard because of all the problems with the European 
layouts, which is why I did not notice this conflict, but I am quite 
sure that it is also present for the Danish layout.

@koenpunt suggested using a modifier with F2, unfortunately option is 
already taken, as I think that would be the preferred binding. Control 
and F2 is a bit more awkward to reach, and semantically we generally use 
option for the “alternative option”.

How about F3 / ⇧F3? The Python bundle does bind “Documentation for 
Module” to F3, so we should change that (I assume Python has an 
alternative documentation command on ⌃H).

infininight:
Not sure the F-keys are a good solution, don't most all people leave them to their default media keys/dashboard/brightness/etc functionality? Think the only free by default are F5 and F6.

koenpunt:
@infininight is that something we have to account for? It's a users choice, and many application use the F keys. I have the F keys always as the F keys and I use the FN key when I want to change my brightness, volume, etc.  IMO every developer should have it set like that, but while saying that, there're probably quite some non-developers that use Textmate. 
Also, when using a non-Apple keyboard, these functions aren't available anyway.

infininight:
Generally defaults are very powerful things, I would suspect that the vast majority* has the default behavior enabled. A difference here might be that marks are a more advanced feature so it might be that the majority for this use does have it disabled.

* Including me, the media/volume keys are just too handy. I never on the other hand use F1-4, though if I had a laptop I suspect I would use brightness more often.

rdwampler:
I think using the F-keys might be the way to go, mainly because we already use F2 and the like for Bookmarks and I think it would allow us to add new "marks" shortcuts in the future in an expected and consistent way. E.g., we could introduce shortcuts (F6 / ⇧F6) to only navigate to warnings/issues/errors and so on.

I agree with @koenpunt. Most users will probably just use the mouse to interact with most marks. If they use Bookmarks often, I imagine they already have changed the default behaviors of the F-keys or they don't mind the additional fn key press.

koenpunt:
> A difference here might be that marks are a more advanced feature so it might be that the majority for this use does have it disabled.


> I imagine they already have changed the default behaviors of the F-keys or they don't mind the additional fn key press.

I think both apply here.

koenpunt:
But then just maybe use <kbd>F3</kbd> and <kbd>SHIFT</kbd><kbd>F3</kbd> to keep it simple.

PR #1317, opened 2016-02-11
Add a couple of explicit link dependencies

rdwampler:
While looking into this [issue](http://lists.macromates.com/textmate/2016-February/039236.html) on the mailing list, I noticed we didn't have an explicit link for the CommitWindow framework. I am not sure if this is an actual problem (or really necessary) as I was unable to reproduce the original error. However, while performing a clean build I encounter a similar issue where it was complaining of `Preferences/Keys.h` not being found when compiling `FindWindowController.mm`. In both case we were missing a linking dependency.

### DIFF
--- a/Applications/commit/target
+++ b/Applications/commit/target
@@ -1,1 +1,2 @@
 SOURCES = src/*.mm
+LINK = CommitWindow
--- a/Frameworks/Find/target
+++ b/Frameworks/Find/target
@@ -2,5 +2,5 @@ SOURCES      = src/*.{cc,mm}
 TESTS        = tests/*.cc
 EXPORT       = src/Find.h src/scan_path.h
 CP_Resources = resources/*
-LINK        += text editor io ns OakAppKit OakFoundation regexp scope
+LINK        += text editor io ns OakAppKit OakFoundation Preferences regexp scope
 FRAMEWORKS   = Cocoa


sorbits:
Sorry I didn’t comment on this one.

The link dependency for the Find framework is correct.

The change for the commit target is slightly more problematic because by adding a link target, we basically link the shell executable with the CommitWindow framework, in practice this doesn’t matter currently because there are no resources in this framework, so it’ll be statically linked, and the optimizer should stirp it, but in theory, it is sort of wrong.

Ideally there should be a way to indicate header dependency.

rdwampler:
> Sorry I didn’t comment on this one.

No worries. I was doubtful that this was a proper fix. Thanks!

> The link dependency for the Find framework is correct.

Just to clarify do you mean the current (master) link dependency for the Find framework is correct or the version in this patch?

sorbits:
> Just to clarify do you mean the current (master) link dependency for the Find framework is correct or the version in this patch?

The version in your patch.

I have now pushed 417193e089485457ad392f86fcb10c7dd2fe5b8f which should fix all link dependencies for all targets (except `OakDebug` framework, but this is not built by default).

I did this using a modified version of `bin/gen_build`, it is something I would like to make the build system automatically handle, but it needs a bit more work than my “hack” since we only want to re-evaluate link dependencies for changed source files (rather than scan all sources and their included headers before each build).

PR #1318, opened 2016-03-23
Update configure: fix `which` Illegal option -s

tiagosiebler:
Trying to compile on Yosemite throws back an error, resulting in dependencies not being recognised, even though they have been installed. This fixes it.

### DIFF
--- a/configure
+++ b/configure
@@ -25,7 +25,7 @@ min_os=10.7
 # = Check if boost is installed =
 # ===============================
 
-if which -s brew && [[ -z "$boostdir" && ! -d /usr/local/include/boost ]]; then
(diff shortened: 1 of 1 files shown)


sorbits:
WHICH(1)                  BSD General Commands Manual                 WHICH(1)
    ⋮
         -s      No output, just return 0 if any of the executables are found, or 1 if none are found.

I think you may have installed a different `which` command that does not implement the BSD options, and have put the install location before `/usr/bin` in your `PATH` variable.

tiagosiebler:
You're right, looks like this version of "which" is a leftover from when I used fink:
```
/sw/bin/which: illegal option -- s
Usage: /sw/bin/which [-a] args
```

PR #1320, opened 2016-04-10
Add a horizontal 2px margin on the Folding scope indicator

orta:
Hey there - long, long time user, first time contributor to TextMate!

I've [noticed](https://github.com/orta/life/issues/68) that when using anything other than `{` and `}` for scope start/ends the type feels quite tight against the folded indicator.

### Before:

screen shot 2016-04-10 at 12 37 02

and

So I've added padding to the file:

screen shot 2016-04-10 at 12 55 29

and adjusted the constant metrics.

### After

screen shot 2016-04-10 at 12 27 12

With 

screen shot 2016-04-10 at 12 26 59

---

I've looked at other merged PRs and I don't see any updates to changelings ( heh, CHANGELOGs), so I think this should be enough - but happy to iterate :+1:


sorbits:
Merged as 34fa11a, thanks!

Changelog is generated (and then edited) from git commit log prior to each release, so no need to update the changes file.

PR #1323, opened 2016-04-27
Select print font size

persquare:
Adding the ability to select font size in the print dialog.

This contribution is Public Domain.


sorbits:
Merged as c5d7aa112d0b3575941f770847c61242b56417b5, thanks!

I merged the two commits into one, also capitalized both words in the label, right-aligned it, and made a few style changes in the code, etc.

It occurred to me that it should probably be a button showing current font name and size with the font dialog showing when pressed.

persquare:
You're quite right. Should have been a button w. font dialog. Have to save it for a rainy day though...

PR #1324, opened 2016-06-08
General menu related changes

rdwampler:
Commit ccb865a is a trivial bug fix .

The rest of the PR is subjective, but I think the changes to the Navigate menu makes it more consistent and distinguishes it from the Go menu with its "Go To File...", etc.

### DIFF
--- a/Applications/TextMate/resources/English.lproj/MainMenu.xib
+++ b/Applications/TextMate/resources/English.lproj/MainMenu.xib
@@ -804,12 +804,12 @@ CA
                     <modifierMask key="keyEquivalentModifierMask"/>
                     <menu key="submenu" title="Navigate" id="422">
                         <items>
-                            <menuItem title="Go to Line…" keyEquivalent="l" id="389">
+                            <menuItem title="Jump to Line…" keyEquivalent="l" id="389">
                                 <connections>
                                     <action selector="orderFrontGoToLinePanel:" target="-1" id="390"/>
                                 </connections>
                             </menuItem>
-                            <menuItem title="Go to Symbol…" keyEquivalent="T" id="556">
+                            <menuItem title="Jump to Symbol…" keyEquivalent="T" id="556">
                                 <connections>
                                     <action selector="showSymbolChooser:" target="-1" id="557"/>
                                 </connections>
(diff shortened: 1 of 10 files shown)


sorbits:
As for disabling _Use Selection for Replace_ and _Jump to Selection_ when there is no selection: Both items have a useful purpose with no selection, the former is for using an empty string for replace and the latter scrolls to where the insertion point is.

We could make the menu item titles dynamic to reflect this.

I am unsure if introducing a local variable for the menu item’s `action` method is meant as a source code improvement or simply a run-time optimization. I’m a bit torn on the former because we add a line of code and by introducing a variable we introduce indirection in the comparison statement. The run-time cost of calling the method a dozen times is probably negligible.

As for the menu reorganization, I do like it, but worth mentioning that the _Jump to Selection_ actually does differ a bit in that it scrolls the text view without moving the insertion point where the (former) _Go …_ items all move the insertion point.

Since the menu item title is a standard item, I wonder if “Jump” is Apple’s non-technical word for scroll, and the _Scroll → Line Up, Down / Column Left, Right_ submenu items should all be “Jump” items? Or perhaps the distinction between moving around by scrolling or via insertion point is not worth spelling out in the menus?

Also worth mentioning that the current _Go_ menu is actually 80% file browser related actions, so perhaps this is what should be renamed?

So in summary, I am OK with merging 8ee98f7 and ac77161 as they are but would like to hear your opinion on my thoughts above.

rdwampler:
> Since the menu item title is a standard item, I wonder if “Jump” is Apple’s non-technical word for scroll, and the Scroll → Line Up, Down / Column Left, Right submenu items should all be “Jump” items? Or perhaps the distinction between moving around by scrolling or via insertion point is not worth spelling out in the menus?

I think "Jump" in all of these cases suggest a possibility of an abrupt/quick change not necessarily moving of the insertion point. Since the Scroll submenu items are all small incremental changes they shouldn't be renamed.

> Also worth mentioning that the current Go menu is actually 80% file browser related actions, so perhaps this is what should be renamed?

I am actually leaning towards renaming the _Go_ menu in addition to the other changes, but I don't know what would be a good name. The _Go To Tab_ item doesn't quite fit amongst the other items which makes naming difficult. Maybe, we can move it to the _Window_ menu (_Select Next/Previous Tab_ are already located there) and rename the _Go_ menu to something more explicit like "File Browser"? idk.

sorbits:
Merged menu changes as 162b3de...976a794.

I think it makes sense to move _Go to Tab ▸_ to _Windows_ menu (_Select Tab ▸_?) then _Go to File…_ can become _Quick Open…_ in the _File_ menu and then we are left with _Go to Related File_ which is probably a bit of a niche feature, technically should probably also be _File_ menu though fear we are adding too much stuff to that menu.

PR #1325, opened 2016-06-14
fixup! Ignore /usr/bin/git if /usr/bin/xcode-select -p fails

rdwampler:
The missing NULL was causing TextMate to crash.


sorbits:
👍

PR #1326, opened 2016-06-14
Xcode 8 beta related fixes

rdwampler:
cf4ac54 is probably just related to my system, but it would be nice to have it upstream.


sorbits:
Merged as 2210264...162b3de, thanks (fixed typo in summary of 745a9eb)!

PR #1327, opened 2016-06-19
Allow gutter fontName to be set independently of editor fontName

meyer:
I like Xcode’s gutter font (Xcode Digits) and I wanted to use the font in TextMate. I wasn’t sure where the best place is for a setting like this, but since the scale factor is already in `NSUserDefaults`, saving the gutter font name there as well made sense to me.

Xcode Digits lives here:
Xcode Digits


sorbits:
Merged as 91befc2, thanks!

I changed `objectForKey:` to `stringForKey:` when looking up the font name in `NSUserDefaults`.

meyer:
@sorbits that was quick, thanks! I just won a tiny bet… my roommate said I should use `objectForKey` since `NSString` is an object, I said I should use `stringForKey`.

PR #1328, opened 2016-06-19
Go menu changes

rdwampler:
This PR implement the suggestions in #1327.

After interacting with it a little bit, I personally think the new menu layout is more intuitive. I moved _Go to Related File_ to the _Navigate_ menu. The motivation was based on Xcode's _Navigate_  → _Jump to Next Counterpart_ which has similar behavior, though I didn't rename it using the "Jump to" prefix since the other "Jump" items work within the current file and also after changing  the name of the _Go_ menu  to _File Browser_, the usage of "Go" seems less of a concern.

 The commits should be atomic so you can pick and choose.

### DIFF
--- a/Applications/TextMate/resources/English.lproj/MainMenu.xib
+++ b/Applications/TextMate/resources/English.lproj/MainMenu.xib
@@ -1038,8 +1038,8 @@ CQ
                         </items>
                     </menu>
                 </menuItem>
-                <menuItem title="Go" id="299">
-                    <menu key="submenu" title="Go" id="300">
+                <menuItem title="File Browser" id="299">
+                    <menu key="submenu" title="File Browser" id="300">
                         <items>
                             <menuItem title="Back" id="545">
                                 <modifierMask key="keyEquivalentModifierMask"/>
--- a/Applications/TextMate/resources/English.lproj/MainMenu.xib
+++ b/Applications/TextMate/resources/English.lproj/MainMenu.xib
@@ -892,6 +892,13 @@ CA
                                 </menu>
                             </menuItem>
                             <menuItem isSeparatorItem="YES" id="804"/>
+                            <menuItem title="Go to Related File" keyEquivalent="" id="303">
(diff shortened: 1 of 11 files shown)


sorbits:
Merged as 91befc2...7e34978. This prompted me to further improve naming of the select/deselect file browser menu items (5c7b04f...0a0b9a8) which I think should make this functionality easier to discover.

With a dedicated file browser menu there are a few other items that might also benefit from being moved to this menu, I am thinking of _New File_ and _New Folder_ presently found in both the _File_ menu and the file browser’s own action menu. With no file browser these items are always disabled.

There is also _View_ → _Show File Browser_ but I am not sure what the convention is here. It would be reasonable to expect a user to look in the _File Browser_ menu for how to “show this thing”.

I’m also thinking the current “Go” items (somewhat mirroring Finder’s _Go_ menu) in the _File Browser_ menu could maybe benefit from being indented with a menu heading like _Go to_.

Your thoughts are appreciated.

rdwampler:
> Merged as 91befc2...7e34978. This prompted me to further improve naming of the select/deselect file browser menu items (5c7b04f...0a0b9a8) which I think should make this functionality easier to discover.
👍 

> With a dedicated file browser menu there are a few other items that might also benefit from being moved to this menu, I am thinking of New File and New Folder presently found in both the File menu and the file browser’s own action menu. With no file browser these items are always disabled.
> 
> There is also View → Show File Browser but I am not sure what the convention is here. It would be reasonable to expect a user to look in the File Browser menu for how to “show this thing”.
> 
> I’m also thinking the current “Go” items (somewhat mirroring Finder’s Go menu) in the File Browser menu could maybe benefit from being indented with a menu heading like Go to.

I am in favor with the above changes, though I think  _View → Show File Browser_ is OK. A quick look at other apps, they mostly seems to put Show/Hide "something" under _View. So in terms of discoverability I think keeping it where it is won't hurt.

PR #1329, opened 2016-06-21
Pass our minimum deployment target to ibtool

rdwampler:
Since we are making changes to the main.xib, I decided to dig into the cause of
```
warning: This file is set to build for a version older than the deployment target. Functionality may be limited. [9]
``` 
and found a few google results about CocoaPods experiencing similar issues and passing `--minimum-deployment-target` to `ibtool` to resolve it. It probably wasn't causing any problems or worth fixing, but it does make a clean build a little less noisy. 😉 Also, as I mentioned in the commit message, I don't think it will be necessary when we move the deployment target to 10.8 or higher.

### DIFF
--- a/bin/gen_build
+++ b/bin/gen_build
@@ -951,7 +951,7 @@ rule dsym
   description = Archive dSYM info for ‘$in’…
 
 rule compile_xib
-  command = xcrun ibtool --errors --warnings --notices --output-format human-readable-text --compile $out $in
+  command = xcrun ibtool --errors --warnings --notices --output-format human-readable-text --minimum-deployment-target $APP_MIN_OS --compile $out $in
   description = Compile xib ‘$in’…
 
 rule process_plist


sorbits:
> It probably wasn't causing any problems or worth fixing

Definitely worth fixing, thanks (9337b12)! 😌 I have previously wasted time with `-mmacosx-version-min`, `MACOSX_DEPLOYMENT_TARGET`, and going through the UI of Xcode, trying to suppress this annoying warning.

Curious why you think it is worth removing if we bump the deployment target (should it no longer show a warning). Seems to me that providing this information (to `ibtool`) will always be useful.

rdwampler:
> Curious why you think it is worth removing if we bump the deployment target (should it no longer show a warning). Seems to me that providing this information (to ibtool) will always be useful.

Yeah, you're right, I was just thinking of the immediate issue of suppressing the warning and default would be 10.9 (but I imagine Apple keeps moving it with every release), so after upgrading the error will go away, but moving forward you're correct that it would be nice to always provide it to trigger mismatch warnings.

PR #1331, opened 2016-06-30
adjust CJK graph spacing to double ascii width

windwiny:
Set TextMate2  use fixed width english fonts, like Monaco, PT Mono ,
One Chinese/Japanese/Korean character graph small than two ascii graph width, 
this patch add two parameters, set them enable expand some character fonts spacing

```
defaults write com.macromates.TextMate.preview g_need_kern_noascii -bool YES
defaults write com.macromates.TextMate.preview g_double_width_unicode -integer 9472
```

this still has there are problem:
i. performance (check each character on line)
ii. emoji character (width bigger than double ascii width)
iii. which unicode range graph need set kern-attribute

Any idea?


sorbits:
For me, Japanese glyphs are roughly twice the width of English characters.

I tried this with both Monaco, PT Mono, and the Menlo font.

So I don’t understand why this commit is necessary. Can you post a screen capture of how text is rendered before/after your change?

windwiny:
when select Ubuntu Mono font, chs glyphs is twice width of ascii character, not do any thing.



when select PT Mono font, will expand chs fonts spacing

sorbits:
Thanks for the clarification. So we want CJK glyphs to be **exactly** the width of two ASCII characters for proper alignment, correct?

I think the problem with PT Mono etc. is that these fonts do not actually have CJK glyphs, so they are taken from other fonts, and thus, do not have the proper width. Unlike Ubuntu Mono, which include the glyphs.

As for your last question about which characters to “fix” then we already have [`text::is_east_asian_width`](https://github.com/textmate/textmate/blob/master/Frameworks/text/src/ctype.cc#L30) that you can use although we should probably update the function to use `CFCharacterSet` for improved performance.

sorbits:
List of double-width characters is from a Unicode publication: 45f847d01ec774ec4a621f553116b9a6b5dff865

windwiny:
@sorbits 
Yes, i want CJK glyphs to be exactly the width of two ASCII characters for proper alignment.


45f847d script only add `Wide` char to east_asian_width charset, i think the `Fullwidth` char also need , like `FF08` `FF09` etc.. [jrgraphix.net/r/Unicode/FF00-FFEF](http://jrgraphix.net/r/Unicode/FF00-FFEF) , ab6946a 75bc56f changed script `utils/gen_eaw.rb` . 
The `Ambiguous` and  `Neutral` looks too complexity, just ignore them. 

maybe the problem with fonts, most emoji width bigger than double ascii width, i create `is_east_asian_width_minus_emoji` , let east_asian_width charset minus a part of emoji chars.

windwiny:
Report an text-relative bug, from [this page](jrgraphix.net/r/Unicode/0900-097F)  copy `0963` character , paste it to new empty file, keep this char on first position, this char cannot delete, input other chars, `ctrl-a` `ctrl-e` hot key not work correct, and some time TM2 will crashed.

sorbits:
Be aware that `CFStringGetCharacterAtIndex` return UTF-16 values (which could be part of a surrogate pair) because `CFStringRef` is backed by UTF-16.

If you saw exceptions (as I see you added a `try`/`catch`) then perhaps that is the reason.

I would avoid iterating a `CFStringRef` and instead work on the source material via the [`utf8::make`](https://github.com/textmate/textmate/blob/master/Frameworks/text/src/utf8.h#L105) helper function to create an iterator that advance one multi-byte character at a time.

This does however require translating from UTF-8 to UTF-16. There is a helper function for this, you can look e.g. [how we handle tab stops](https://github.com/textmate/textmate/blob/master/Frameworks/layout/src/ct.cc#L207) but another approach is to build up the target string gradually, which is what is done while we style the string.

windwiny:
Sorry, i will clarification, i'm mean paste 0x0963 char to empty file first position , will cannot delete it,
test on `TextMate_2.0-beta.11.3` have some problem, it's not this patch reason. 

Another, this patch's try/catch block not catched any exception on my those days test.

sorbits:
Thanks, I did understand the bug report, my comment was not related to that. Try <kbd>⌘R</kbd> the following code and I think you will see what I mean:

```cpp
#include <CoreFoundation/CoreFoundation.h>

int main (int argc, char const* argv[])
{
	CFStringRef str = CFSTR("🤐");
	for(CFIndex i = 0; i < CFStringGetLength(str); ++i)
		fprintf(stderr, "%2ld: %08X\n", i, CFStringGetCharacterAtIndex(str, i));
	return 0;
}
```

Ideally the output would be:

	 0: 0001F910

But actual output is:

	 0: 0000D83E
	 1: 0000DD10

windwiny:
Looks to change fonts show style is very difficult, and add glyphs spacing seem not good. 😞

Maybe TM can provide some feature:
add some predefined font style;
Comments in the program code/ Markdown's code block/ table  can use `real fixed width` fonts;
font info can saved on file's extended attributes, just like current position, selected text etc.. 

@sorbits  how do you say?

sorbits:
> font info can saved on file's extended attributes, just like current position, selected text etc

I do not understand this part. We have to store selection etc. as extended attributes if we want to restore them.

For _double width fonts_ would you let the user select what parts they want as _double width_ (and store in extended attributes)?

Or would you just like to limit the _double width_ processing to comments, code blocks, etc.? For the latter, the user can already set a font based on scope, for example for Markdown headings we change the font to _Baskerville_. You could make a similar (scoped) setting for `comment` and change the font to _Ubuntu Mono_.

But I assume you would prefer to still use whatever is your current font, only for CJK glyphs would you want _Ubuntu Mono_ or alternatively, that TextMate change the spacing of the chosen font so that glyphs are exactly the width of two regular (latin) characters.

To me it seems the ideal solution would be to fix the font (to have CJK glyphs that are double width), and the less ideal solution would be for TextMate to enforce this for all CJK glyphs, if the user enables the setting. I do not understand why you would only apply the _double width_ enforcement on subsets of the document. Are you suggesting this only because applying it to the full document causes issues, i.e. not all characters identified in the Unicode East Asian Width should really be forced as _double width_, or possibly, some characters not in this list, should have _double width_.

sorbits:
Short version: Why don’t we want to force _double width_ for all CJK characters in the document?

PR #1332, opened 2016-07-04
Use thread_local instead of our own implementation

rdwampler:
Apple's clang (Apple LLVM version 8.0.0 (clang-800.0.24.1)) as shipped with Xcode 8 now supports `thread_local`. This effectively reverts commits: 14c6a70 f3d1aba 9d1d1a1a and 172ce9d.

---
Although Xcode 8 is still in beta and will be for some time, I wanted to go ahead and submit this PR as a reminder. Also, my fork receive most traffic related to this issue so it may help others if they are searching about `thread_local` support with clang.


sorbits:
Merged as 1658d6356a0e8d341e769234271505bdc977119e now that Xcode 8 is out of beta, thanks!

PR #1334, opened 2016-07-06
Use non-deprecated method to load nib (10.8)

rdwampler:
We can pass `NULL` to `topLevelObjects` since we have a strong IBOutlet to the NSPanel (our top-level object).


sorbits:
Merged as 69b7b8ee42c633965efee44bba2efaaa5e77347b, thanks as always!

FYI The number of TextMate users on 10.8 seems to be in the hundreds, so I am thinking one more regular release, and then we can drop support for 10.8 as well.

PR #1335, opened 2016-07-13
Refactor ODBEditorSuite to avoid FSRef API (10.8)

rdwampler:
I have tested this using MarsEdit and it appears to be working correctly, though you may want to scrutinize it.

I attend to take care of the other usage of the FSRef API; however, that appears to be working around a bug in 10.8 where the deleting a symbolic link would delete the actual file and not the link. I believe this was resolve in 10.8.3, so maybe hold off on removing it until we go to 10.9.

### DIFF
--- a/Applications/TextMate/src/ODBEditorSuite.cc
+++ b/Applications/TextMate/src/ODBEditorSuite.cc
@@ -1,4 +1,5 @@
 #include "ODBEditorSuite.h"
+#include <cf/cf.h>
 #include <document/collection.h>
 #include <oak/debug.h>
 #include <text/hexdump.h>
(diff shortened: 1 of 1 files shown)


sorbits:
Much appreciated, merged as fa9ceead091cbbfa7c69914a0b1e0443e23359dc.

The problem with `typeFileURL` (as best as I recall) was with third party software that did not recognize the file sent back unless I used `typeFSRef`.

This could possibly have been because the file was in `/tmp` which when resolved became `/private/tmp` or maybe it had to do with changed inodes when using atomic save, and the application using this (“catalog ID”) to track open files.

Anyway, this is 10+ years ago and I have forgotten which third party software had the issue plus there is the `keySenderToken` to allow applications to track what files they have open, so I see no need for preserving this “workaround” in light of the `FSRef` deprecation.

I tested your change myself with CyberDuck and also had no problems, I will add a note in the release notes for people to report any issues with third party software using TextMate as external editor.

PR #1336, opened 2016-07-13
Manually release GCD objects in C++ libraries

rdwampler:
Since these are pure C++ libraries, they are not managed by ARC.


sorbits:
Ups…😊 

Merged as e50dc57db04bb5426be6db6990742a48518e2332, thanks as always!

PR #1338, opened 2016-08-11
README: tweak build instructions.

MikeMcQuaid:
Make `./configure` install the packages it needs using `brew bundle` from a `Brewfile`, setup submodules and default to the version of OS X that you're currently on (to avoid warnings) but allow it to be configurable.

Also, update Travis CI's configuration and the README to simplify things given this new configuration.

This should make it easier for new contributors to get started. Happy to tweak this PR as desired. This patch should be considered public domain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,12 +30,12 @@ To bootstrap the build you need to run `./configure` (in the root of the source 
 
 In the simplest case you would run:
 
-	git clone https://github.com/textmate/textmate.git
+	brew install ragel boost multimarkdown hg ninja capnp google-sparsehash libressl
+	git clone --recursive https://github.com/textmate/textmate.git
 	cd textmate
-	git submodule update --init
 	./configure && ninja
 
-Please note that if you downloaded the source code (rather than cloned via git) you likely miss the submodules and the build will therefore fail.
+Please note that if you downloaded the source code zipball/tarball (rather than cloned via git) you miss the submodules and the build will therefore fail.
 
 ## Prerequisites
 
@@ -50,7 +50,11 @@ To build the source the following must first be installed on your system:
  * [Cap’n Proto][capnp] — serialization library
  * [LibreSSL][libressl] - OpenBSD fork of OpenSSL
 
-You need to manually install [Cap’n Proto][capnp] if you're not using [homebrew][]. To install the other dependencies via [MacPorts][] run:
+To install these dependencies using [Homebrew][] run:
+
+	brew install ragel boost multimarkdown hg ninja capnp google-sparsehash libressl
+
+You need to manually install [Cap’n Proto][capnp] if you're not using [Homebrew][]. To install the other dependencies via [MacPorts][] run:
 
 	sudo port install ninja ragel boost multimarkdown mercurial sparsehash libressl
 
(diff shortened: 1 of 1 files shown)


sorbits:
I don’t think `./configure` should do things like update git submodules (what if we want to build and older version and explicitly checkout old submodules first?) nor install dependencies outside the `build` directory (we may have `brew` installed but prefer dependencies installed from elsewhere, e.g. directly from source or with build flags that `brew` cannot handle).

I am also not sure building for current OS as minimum requirement is a good idea because this may actually fail, for example when we updated the minimum requirement to 10.8 then we had to remove `dispatch_release` functions from Objective-C code, as that is now handled by ARC and would result in double release if the code was not updated.

MikeMcQuaid:
> I don’t think ./configure should do things like update git submodules (what if we want to build an older version and explicitly checkout old submodules first?)

How about if this was run only if e.g. none of the submodules exist/have been initialised?

> nor install dependencies outside the build directory (we may have brew installed but prefer dependencies installed from elsewhere, e.g. directly from source or with build flags that brew cannot handle).

I assume this was the intent given you're using `brew --prefix` to look up the include paths. How about if the instructions to `brew install` are instead changed to `brew bundle`? That would also simplify CI setup.

> I am also not sure building for current OS as minimum requirement is a good idea because this may actually fail, for example when we updated the minimum requirement to 10.8 then we had to remove dispatch_release functions from Objective-C code, as that is now handled by ARC and would result in double release if the code was not updated.

Ok, will change the default back to 10.8. Any objection to leaving it as a configurable variable?

rdwampler:
How about creating a bootstrap script and putting these changes there?

sorbits:
A bootstrap script does sound more appropriate for these things, but I am a little torn on the necessity for this. We can’t avoid the `README`, and the current steps can on a system with homebrew can be reduced to:

    brew install ragel boost multimarkdown hg ninja capnp google-sparsehash libressl
    git clone --recursive https://github.com/textmate/textmate
    cd textmate && ./configure && ninja

If we go with the `brew bundle` system and a bootstrap script, we end up with *two* new files, but the user would still need to basically copy/paste the same lines from the `README`.

As for making the minimum OS requirement configurable, I don’t understand the motivation for this. As mentioned above, the SDK can’t always be changed independently of the source.

P.S. I wouldn’t mind updating the `README` to be more biased toward homebrew and put the above instructions near the top and then elaborate about dependencies later in the `README`.

MikeMcQuaid:
> git clone --recursive https://github.com/textmate/textmate

@sorbits This seems like a good idea; mind if I add this to the README?

> As for making the minimum OS requirement configurable, I don’t understand the motivation for this. As mentioned above, the SDK can’t always be changed independently of the source.

Mainly the opposite of the suggested LibreSSL changes: to make the warnings shut up. I'll happily drop it, though.

> If we go with the brew bundle system and a bootstrap script, we end up with two new files, but the user would still need to basically copy/paste the same lines from the README.

My thinking with using `brew bundle` is that it's a nicer (and quicker) way that telling people to install a list of things, some of which they may already have installed. It also reduces duplication between the README and .travis.yml if dependencies change. That said, happy to drop it if you think that's best.

> P.S. I wouldn’t mind updating the README to be more biased toward homebrew and put the above instructions near the top and then elaborate about dependencies later in the README.

Makes sense to me.

--

@sorbits Happy to go with whatever you decide is best based on the above. Will wait for reply to this comment and then implement whatever you think is best. Thanks!

sorbits:
You’re more than welcome to improve the `README`.

The LibreSSL changes are for expert users and not “suggested” per 
se. But LibreSSL should not be tied to a specific version of the Mac OS 
X SDK (as it is not using AppKit / FoundationKit nor is written in 
Objective-C / using ARC), very different from TextMate.

As for duplication, that is true, but there already is some duplication 
between homebrew and MacPorts instructions, and previously we had more 
dependencies for 10.7 users, which I don’t think the brewfile would 
have been able to handle. I don’t see myself adding more dependencies 
though, on the contrary, I would like to get rid of some of the current.

MikeMcQuaid:
> You’re more than welcome to improve the `README`.

Done, the PR now focuses just on README changes.

> and previously we had more
> dependencies for 10.7 users, which I don’t think the brewfile would
> have been able to handle

Off-topic at this point but you can do anything you can do with Ruby in the `Brewfile` e.g. conditionals based on OS versions or `system` command results.

sorbits:
Thanks, I amended your changes with a more thorough overhaul of the file so I didn’t preserve the commit author but did mention you in the summary, I hope that is acceptable 😄

MikeMcQuaid:
@sorbits Great, thanks!

PR #1339, opened 2016-08-17
Add attr.project.swift scope

rdwampler:
This scope can be used to support the Swift Package Manager tools (e.g., `swift build`).

-

I am not sure if the build command should be added to the official Swift tmbundle since it is technically part of the Swift Package Manager (Swift 3), which contains a few other tools, so I created a separate tmbundle [swift-package-manager.tmbundle](https://github.com/rdwampler/swift-package-manager.tmbundle). Currently, it only supports building a Swift package, though I plan to add more commands to support `swift package` and `swift test`.

Let me know if you prefer otherwise and I will submit a PR to the official Swift tmbundle.

### DIFF
--- a/Frameworks/DocumentWindow/src/DocumentController.mm
+++ b/Frameworks/DocumentWindow/src/DocumentController.mm
@@ -2070,7 +2070,7 @@ namespace
 - (void)setCommandRunner:(command::runner_ptr const&)aRunner
 {
 	_runner = aRunner;
-	if(self.htmlOutputInWindow || self.htmlOutputView.runningCommand)
+	if(self.htmlOutputInWindow || self.htmlOutputView.isRunningCommand)
 	{
 		HTMLOutputWindowController* target = nil;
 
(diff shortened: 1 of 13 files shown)


sorbits:
Merged as 860da518cb58e61f58e677c84fc4238d401fe8aa, thanks!

As for the Swift Package Manager bundle, I don’t have any real thoughts on the matter currently, as I am unfamiliar with this tool suite.

It probably does seem proper to have it be its own bundle since it has a “Build” command which is not about building swift projects (I assume).

Btw: I do intend to extend the current bundle suggestions to also suggest “project” bundles, so discoverability (for users with a `Package.swift` file) should improve.

rdwampler:
> It probably does seem proper to have it be its own bundle since it has a “Build” command which is not about building swift projects (I assume).

The focus for now is mainly on building/packaging swift modules/libraries; however, recently it can be use to build C, C++ and Objective-C++ projects (before one would had to build C libraries separately, since it could only link with them).

PR #1340, opened 2016-08-31
Visual improvement to OakChoiceMenu when using large fonts

rdwampler:
On a related note, I would like to apply these changes to the `DIALOG` popup (along with some other enhancements), but since there is already some duplication of code between TextMate and Dialog, I would like to move the Dialog plugin into the main source tree in order to reduce the duplication and easily share more of the code between the two. Thoughts?


sorbits:
Merged as 89d45f7...f59c445, thanks!

I’ve considered making Dialog part of the main source tree as well, as git submodules suck.

To be honest, the main thing holding me back is the code quality, with the submodule, it’s an isolated mess that is easier for me to ignore :smile: but I guess you could argue that because it’s a submodule, it’s also less likely to be cleaned up.

PR #1341, opened 2016-08-31
Use Core Animation for find indicator

jtbandes:
Animations are now more smooth, and more visually similar to NSTextView `showFindIndicatorForRange:`. The code is simpler too. It's hard to see the difference in GIFs, but here you go:

Before:

After:


sorbits:
I get this error:

    Frameworks/OakAppKit/src/OakPopOutAnimation.mm:13:36: error: cannot find protocol declaration for 'CAAnimationDelegate'; did you mean 'NSAnimationDelegate'?

It looks to be a macOS 10.12 symbol, but simply removing the protocol conformance declaration fixes it (no warnings).

sorbits:
Merged as ac6d144a6e30a1c1b014304cda2fc15964ada56f, thanks!

PR #1342, opened 2016-08-31
Don't store our font scale factor as a percentage

rdwampler:
While we want to show the font scale factor as a percentage in the UI, internally, we should represent it as a CGFloat, which is more consistent with the NSFont APIs and makes interacting with them easier.

--

Just a nitpick. 😄


sorbits:
I have a strong preference for integers especially when basically doing integer math (as we do when adding and subtracting a step factor and comparing them against the lower bound).

Nonetheless I merged this as 06ba9b6dab0b98f5ab029b963c19de4ba05bdb69 because the scale factor is sort of public API for `OakTextView`, so it would make sense to use the most natural representation, at least in the API.

PR #1344, opened 2016-09-01
Enable bundle editor window’s document icon

jtbandes:
When there is only one path for an item, that path is used as the representedURL. When there are multiple paths, we still allow showing the menu, but it has one item per path instead of showing the directory tree.


sorbits:
Merged as 1e440b3f99f3abce25ddd50ff16fed049441ff69, thanks!

PR #1347, opened 2016-09-09
Don't set our document with a default constructed document_ptr

rdwampler:
This was never necessary since our documentView will set this to
document_ptr() when it's released, though it was harmless. But now,
document_ptr is just a thin wrapper around OakDocument and the default
constructor does nothing. As a result we were basically setting the
text view OakDocument instance to nil.

After commit f9e02d9, when the commit window closes, this would
noticeably cause the text view to turn "white" when using a dark theme.


sorbits:
Merged as 0c71b54a9ea33827485dc96fd9811d76b1c42839, thanks!

PR #1348, opened 2016-09-10
Add “git config commit.gpgsign false” to SCM tests

jtbandes:
This allows tests to pass even when a global .gitconfig has `gpgsign=true`.


sorbits:
Merged as 7ec00a5bea31466de7494a21dcd00b829c15b61d, thanks!

PR #1349, opened 2016-09-10
Use yellow background and underline for project find results

jtbandes:
This style is a de facto standard across a handful of popular Mac apps.

screen shot 2016-09-10 at 12 50 09 am

### DIFF
--- a/Frameworks/CrashReporter/src/CrashReporter.mm
+++ b/Frameworks/CrashReporter/src/CrashReporter.mm
@@ -151,20 +151,17 @@ static std::string create_gzipped_file (std::string const& path)
 				{
 					didSend.insert(report);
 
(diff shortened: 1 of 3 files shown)


sorbits:
Merged as e0a49637f98df8b5b272037441ad4cc4d2d5b3b0, thanks!

How did you settle on these exact colors?

jtbandes:
I inspected some popular implementations of this style, used a color meter, and did a few calculations 😉

jtbandes:
(of course, there is no absolute standard to uphold — change them to your liking!)

rdwampler:
Should this style be applied to `OakChooser`'s "search" results too? (Xcode seems to apply this everywhere for all "search" results)

jtbandes:
@rdwampler That's a good point, it probably should be.

PR #1351, opened 2016-09-10
Highlight the selection again when it’s the only find result

jtbandes:
This more closely matches the behavior of NSTextView.


sorbits:
Merged as 73330e29db70153dc09a1aa0fffb72b01d24be8f, thanks!

PR #1356, opened 2016-09-12
fixup! Only set table view’s row height during sizeToFit

rdwampler:
We need to set the row height after calling sizeToFit on textField.

---

I initially had the same code, but at larger font scales the table height would not be correct, so I set it in the delegate method.  I didn't notice any issues so I never revisited it.


sorbits:
Thanks, the reason I did it before sending `NSTextField` an explicit `sizeToFit` was because my mind was in the `OakPasteboardSelector` code where we can have multiline rows, and wrongly assumed, that initializing an `NSTextField` with a font would give it proper initial dimensions.

PR #1359, opened 2016-09-13
Let OakPopOutAnimation set the color in the image to black

rdwampler:
I'm not sure if this is the best way to achieve this, @jtbandes do you have any suggestions?

### DIFF
--- a/Frameworks/OakAppKit/src/OakPopOutAnimation.mm
+++ b/Frameworks/OakAppKit/src/OakPopOutAnimation.mm
@@ -53,6 +53,12 @@ void OakShowPopOutAnimation (NSRect popOutRect, NSImage* anImage)
 
 	OakPopOutView* aView = [[OakPopOutView alloc] initWithFrame:[window contentView].bounds popOutRect:popOutRect];
 	[aView setAutoresizingMask:NSViewWidthSizable|NSViewHeightSizable];
+
+	[anImage lockFocus];
+	[[NSColor blackColor] set];
+	NSRectFillUsingOperation((NSRect){ NSZeroPoint, [anImage size]}, NSCompositeSourceAtop);
+	[anImage unlockFocus];
+
 	aView.contentImage = anImage;
 	[[window contentView] addSubview:aView];
 


jtbandes:
I'm not sure what this would do to subpixel antialiasing — it might be better to just use black when rendering the text in the first place.

But, I imagine @sorbits intentionally allowed color? Having color is a bit different from NSTextView, but I think it's kinda nice (although there are visibility problems when text is white).

sorbits:
I won’t insist on the color, I do like it very much for bright backgrounds, but it’s definitely a problem for dark backgrounds.

Rather than use the `imageForRanges:` (internal) API we could use `attributedSubstringForProposedRange:` and then change the string color to black. This is used for the dictionary pop-up (<kbd>⌃⌘D</kbd>) and looks decent, this might also solve the “clipping problem” that @jtbandes has been concerned with :)

Minor issue with the `attributedSubstringForProposedRange:` API is that it’s using UTF-16 ranges, should probably split that method into two, one that takes UTF-8 ranges, and one that translates from UTF-16 to UTF-8 and then calls the “internal” API (the UTF-16 method is there for the `NSTextInputClient` protocol or somesuch).

Did I just make this a whole lot more complicated?

rdwampler:
> Did I just make this a whole lot more complicated?

Yes! But it's better than trying to manipulate the image.😄 Let me see how that works. Thanks!

rdwampler:
I try this, but `attributedSubstringForProposedRange:` does not handle a multi-line search string. I originally wanted to avoid having `layout_t` know whether or not it is rendering for our text view or for the `OakPopOutAnimation`  Perphaps one approach is to add the ability to disable setting color in `layout_t` and let `OakPopOutAnimation` manage a view with its own copy of a `layout_t` and then have it to decide to disable the color. Idk. Or would it be possible to add a scope like dyn.selection and let the theme handle how to render the matched result since it only a problem for dark themes?

sorbits:
I have merged your original solution as f38eaa5315faf187dee1634359688d8f9d9c7aeb.

I didn’t consider the newlines, thanks for looking into this!

Maintaining a parallel `layout_t` is probably overkill, I’ve added a note to my `layout_t` section about the need to render with “disabled” colors.

PR #1361, opened 2016-10-10
Opt out of automatic window tabbing (10.12)

rdwampler:
Apple recommends applications to respect the system wide users settings for tabbing, but I haven't looked into what this will take. More information is [here](https://developer.apple.com/library/prerelease/content/releasenotes/AppKit/RN-AppKit/index.html#10_12Window%20Tabbing).

### DIFF
--- a/Applications/TextMate/src/AppController.mm
+++ b/Applications/TextMate/src/AppController.mm
@@ -33,6 +33,12 @@
 #import <scm/scm.h>
 #import <text/types.h>
 
+#if !defined(MAC_OS_X_VERSION_10_12) || (MAC_OS_X_VERSION_MAX_ALLOWED < MAC_OS_X_VERSION_10_12)
+@interface NSWindow (Sierra)
+@property(class) BOOL allowsAutomaticWindowTabbing;
+@end
+#endif
+
 OAK_DEBUG_VAR(AppController);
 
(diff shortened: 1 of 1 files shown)


sorbits:
Merged as ec9bd3a002a39a7b4518ea83001b08710d710ee8.

I had to downgrade to Xcode 7.3.1 because the builds from Xcode 8 were unstable (for some users) on macOS 10.12 (only), I suspect it’s related to ARC. Since Xcode 7.3.1 does not have class property syntax, I changed your commit to use `setAllowsAutomaticWindowTabbing:`.

As for respecting the system wide users settings, we may need a hybrid, because I think most users would not want new windows from files opened via Open Quickly…, file browser, or folder search, yet the system settings does not properly support this notion of “related files”.

rdwampler:
I recalled that you were compiling with Xcode 7.3, which is why I added the extension for the 10.11 SDK, but the introduction of the class property in Xcode 8 didn't cross my mind. 😞

> As for respecting the system wide users settings, we may need a hybrid, because I think most users would not want new windows from files opened via Open Quickly…, file browser, or folder search, yet the system settings does not properly support this notion of “related files”.

Yeah, I agree. Probably should also see how Apple's application handle the menu items based on that setting. E.g., does ⌘N change titles based on the setting?

PR #1362, opened 2016-10-16
Change slow string.find method with string.compare method, which is f…

mathbunnyru:
Change slow string.find method with string.compare method, which is faster (analogue of startswith in these cases)

### DIFF
--- a/Applications/mate/src/mate.cc
+++ b/Applications/mate/src/mate.cc
@@ -422,7 +422,7 @@ int main (int argc, char const* argv[])
 
 			if(!files.empty())
 			{
-				if(files[i % files.size()].find(kUUIDPrefix) == 0)
+				if(files[i % files.size()].compare(0, kUUIDPrefix.size(), kUUIDPrefix) == 0)
 						write_key_pair(fd, "uuid", files[i % files.size()].substr(kUUIDPrefix.size()));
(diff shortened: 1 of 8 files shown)


mathbunnyru:
First additional option here - to write a small function, like startswith in Python. But that has some disadvantages - const char* will be converted to std::string, which is an overhead, or we need to write several functions with a lot of copy-paste, which doesn't seem to be a good idea. Another option is to stick to Boost, but in this case, that's overkill and doesn't add anything good. Sorry for my English, btw.

sorbits:
Thanks for the PR, I think for the `const char*` case I would prefer something like `oak::has_prefix` because of the hardcoded string lengths in the code, which I do not like. This could go into `oak/algorithm.h`.

We do currently depend on boost, but only for crc32 and `boost::variant`, the latter soon to be replaced by `std::variant` (C++17) and the former could probably just use sha1 from `CommonCrypto.h`, so probably would be best not to add new boost dependencies (as we may soon be able to cut that dependency).

mathbunnyru:
@sorbits I think you can change my small PR as you want (like adding sth to oak/algorithm.h), because I've just downloaded source code and I'm new to your project (and big c++ projects, to), that's why it's easier for you to make these changes. Thank you for your awesome app!

PR #1363, opened 2016-10-17
Change slow string.find method with string.compare method, which is f…

mathbunnyru:
…aster (analogue of startswith in these cases) and sometimes with oak::has_prefix

I decided not to be lazy and make changes, which you asked. Now for const char *, I use oak::has_prefix.

### DIFF
--- a/Applications/mate/src/mate.cc
+++ b/Applications/mate/src/mate.cc
@@ -422,7 +422,7 @@ int main (int argc, char const* argv[])
 
 			if(!files.empty())
 			{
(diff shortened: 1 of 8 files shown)


sorbits:
I merged the “simple” `find` → `compare` changes as 0b034f9a9dccd9d024d84a6ee0d0e30135bb7534 and then I changed the rest using different methods in 0b034f9...67f5670 (I did not like the extra complexity introduced by using the existing `has_prefix`).

PR #1365, opened 2016-10-18
Replace deprecated `std::random_shuffle`

rdwampler:


mathbunnyru:
@rdwampler I think it would be better to name a new function oak::shuffle instead of oak::random_shuffle. It makes more sense - you clearly show, that you don't use sth. deprecated.

sorbits:
Merged as 1b4adf2...4a7e774, thanks!

@mathbunnyru Since this is API compatible with the old version, I think it’s appropriate to use the name of the old API.

PR #1368, opened 2016-10-20
Changed from NULL to nullptr everywhere in .cc to avoid mistakes like…

mathbunnyru:
… &NULL, casts to ints and so on

@sorbits please, review carefully. I did everything possible to avoid mistakes, but one more check would be nice. I made a change from NULL to nullptr everywhere (in C++ files). That seems to work without any additional changes. Using nullptr instead of NULL is a good idea because of several things like additional error checking during compilation.

### DIFF
--- a/Applications/PrivilegedTool/src/install.cc
+++ b/Applications/PrivilegedTool/src/install.cc
@@ -46,7 +46,7 @@ static void launch_control (char const* command, std::string const& argument)
 	pid_t pid = vfork();
 	if(pid == 0)
 	{
-		execl("/bin/launchctl", "/bin/launchctl", command, argument.c_str(), NULL);
+		execl("/bin/launchctl", "/bin/launchctl", command, argument.c_str(), nullptr);
 		perror("execl(\"/bin/launchctl\")");
 		_exit(EXIT_FAILURE);
 	}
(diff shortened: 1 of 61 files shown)


sorbits:
Merged as 440414f96c3766d46cdad708f0529bfac65a9bf6, thanks!

I changed a few `NULL`’s to the more appropriate `kCFAllocatorDefault` (f78a3b2ce4bbcdcb9ea21473eae29f88bc056aa1).

I also reworded your commit message as the (first) summary line should be less than 72 characters.

mathbunnyru:
@sorbits please, answer this question, if you can.

I try to build using llvm (and clang) which comes from Homebrew and it crashes.

Steps
1. brew install llvm. Current version is 3.9.
2. brew link llvm
3. Change configure script from "xcrun clang" to just "clang" and the same for clang++
4. ./configure and after that ninja

OS X: macOS Sierra
TextMate: current master

Error I get:
test_regexp: 1 of 41 tests failed:
/Users/mathbunnyru/Documents/textmate/Frameworks/regexp/tests/t_match.cc:8: Expected (match[1] == "foo"), found (" != "foo")

I tried to check the error with std::cout, but everything is alright - it prints "foo". Maybe there's an error in macros?

sorbits:
Sorry, I have no idea about why it would fail.

Can you build with Xcode’s clang?

Originally TextMate was not built with Xcode’s clang, and I think 
these macros have remained unchanged since, so this seems a bit weird 
(back then though I used a svn checkout of llvm/clang to get the latest 
version).

PR #1369, opened 2016-10-28
Change from deprecated CFPropertyListWriteToStream

mathbunnyru:
to modern CFPropertyListWrite.
@sorbits I'm gonna make more changes like this if I can to make build more readable (no warnings about deprecated code). But I'm a beginner in Objective-C.


sorbits:
Thanks, merged as 926f7028285170076169a1acf78b9ff1d88db686! I shortened the commit message so that it will not show up truncated in the various contexts where commit messages are shown.

Currently TextMate is built with 10.8 as minimum deployment target, and since this function was not deprecated until 10.10, it should not have caused any warnings (unless you chagned minimum deployment target).

That said, I do of course appreciate patches that get rid of using deprecated API, as I’ll soon drop support for 10.8 and possibly also 10.9.

PR #1372, opened 2016-11-05
Change build_dir to current folder

mathbunnyru:
I think it's better to make build folder as current one (to not mess with ~/ directly).

### DIFF
--- a/configure
+++ b/configure
@@ -6,7 +6,7 @@ function error () { printf >&2 "%s\n\nPlease see README.md for build instruction
 # = Fallback build directory and signing identity =
 # =================================================
 
-: ${builddir:=$HOME/build/TextMate}
+: ${builddir:=./build/}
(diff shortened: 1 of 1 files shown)


sorbits:
I disagree since then the user has to exclude `build` from version control. folder search, open quickly… (⌘T), etc.

If you have a lot of projects, it’s also easier to cleanup temporary (build) files when it’s all in one place.

mathbunnyru:
@sorbits thanks for your reply. Can you say me please, which IDE do you use to develop text mate? I hate, that I can't use between-files jumping, for example.

sorbits:
I use TextMate. Jumping between files? Combination of ⌘T, ⇧⌘F 
(followed by ⌘1-n), and ⇧⌘D (on import/include statements) — but 
my knowledge of the code base might make this easier for me.

PR #1374, opened 2016-11-22
Set minimum OS to 10.9 and address deprecation warnings

rdwampler:
This should address all of the deprecation warnings when you decide to not support 10.8. I also updated the code to use NSAlert instead of the functional APIs, since these will cause a lot noise when building for 10.10. (Feel free to squash the NSAlert related commits–I broke them out to make the review easier). Hope this helps.


sorbits:
Merged as 5f7d5b53d...d1ecba262, thank you very much for this work!

I updated the `NSAlert` commits to use the new (10.12) enumeration names for `alertStyle` and also changed the `NSRunCriticalAlertPanel` introduced in 6e28cea42.

PR #1375, opened 2016-11-29
Ensure file is on disk before adding it to the Recent menu

rdwampler:


sorbits:
We cannot make `isOnDisk` lazy because the proxy icon is dependent on this property via bindings (proxy item appear/disappear to match the document’s “on disk” status).

I assume the commit is because the property is out-of-sync/delayed when you want to test before adding to recent menu?

rdwampler:
>I assume the commit is because the property is out-of-sync/delayed when you want to test before adding to recent menu?

Oops, it was because of line 988 (`self.onDisk        = self.path ? YES : NO;`) afterward we add to the Recent menu. Since this is called after saving I guess there is no reason for the stronger check. The bug is still fix when this commit is dropped.

sorbits:
Thanks, I have merged the latter commit as a499a1559d4b3a7ccf4a53e9f478f97a66fbc233.

PR #1377, opened 2016-12-06
Added the ability to customize the tab title display via new 'tabTitle' tm_properties variable

jakecast:
I updated the OakTabBarViewDataSource tab title method of the DocumentWindowController to use a new setting 'tabTitle' within the .tm_properties file to allow users to change the display name of tabs.

Also moved logic for generating window and tab titles to a new method within the DocumentWindowController called:
(NSString*)titleForDocument:(OakDocument*)document withSetting:(std::string)setting
This was done to remove redundant logic between the window and tab title calls.
Further optimization could be done by storing the docDirectory and settings_t values for each document to avoid generating them twice.

### DIFF
--- a/.tm_properties
+++ b/.tm_properties
@@ -2,6 +2,7 @@
 windowTitleProject   = '${projectDirectory:+ — ${projectDirectory/^.*\///}}'
 windowTitleFramework = '${TM_DIRECTORY/.*\/Frameworks\/([^\/]+)\/.*|.*/${1:+ ($1)}/}'
 windowTitle          = '$TM_DISPLAYNAME$windowTitleFramework$windowTitleProject$windowTitleSCM'
+tabTitle             = '$TM_DISPLAYNAME'
 
 excludeDirectoriesInFileChooser  = '{bin/CxxTest,build}'
 excludeDirectoriesInFolderSearch = '{bin/CxxTest,build}'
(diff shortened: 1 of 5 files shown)


sorbits:
What value would you expect the user to set for `tabTitle`?

I ask because two users have previously requested custom tab titles, but one wanted disambiguation for when multiple open documents had the same name (like `index.html`), and another wanted to “shorten” the display names to allow for more visible tabs.

jakecast:
I have the default tm_properties just showing the current filename `$TM_DISPLAYNAME` to remain consistent, but my personal one I have set up to show the file's location within the current project.
```
tabTitle = ${TM_DIRECTORY:+${TM_DIRECTORY/^${projectDirectory}(?:\/(.+))?/${1:+$1/}/}}$TM_DISPLAYNAME
```

It is handy on certain projects for me where the directory structure dictates the file's purpose..
screen shot 2016-12-16 at 08 30 54

It makes it a lot easier to navigate between open files when too of the files have the same name and their location in the project determines for what they are used..

koenpunt:
I'm one of these users(disambiguation)  and I think this would work rather fine for me :)

jaredreich:
Excellent work, really looking forward to this!

ygbr:
For me this is also extremely useful. Sometimes you do need disambiguation of file names by looking at the full path of files, specially when using web frameworks that have some default files with the same name on different controller / view directories.

Also extremely handy when you're into DevOps and need to deal with a lot of configuration files that have the same name and different paths.

sorbits:
Merged as d380e7ab0c3fd2243f2594c6ec13934108fc245a, thanks, and sorry about the late merge.

Minor changes made to your patch, should not be necessary to check if document has a display name, for the window title, the previous check is really a check to see if there is a document.

jaredreich:
Awesome! When will this be bundled into a release?

PR #1378, opened 2016-12-14
Don't try to replace a folder with an item it contains

rdwampler:
I just ran into this by accident and it resulted in the entire contents of the folder, including the item, to be deleted.

Feel free to change the message text.


sorbits:
Merged as 6e28cea42c364ccda196cb216f8318a55adbbcdb, thanks, and I hope you did not lose data on this!

I thought my table view drag’n’drop code had already a check to disallow this, but if it does, it does not work :( have added a note about looking further into this.

PR #1381, opened 2017-02-26
Treat *.target files as source.tm-properties

mkhl:


sorbits:
Merged as a00f13f071ae2463f09680083e9e3ef21b2daf6e, thanks!

PR #1382, opened 2017-04-12
Resolve ambiguous autolayout issues

rdwampler:
I came across these while trying to resolve some layout issues for something else I'm working on.  I think most, if not all, only occurred after 10.12 (I don't remember seeing any layout issues before) so it maybe a bug in autolayout. But, the number of ambiguous layouts made it hard to debug in `lldb` so I think it's worth fixing despite the additional verbosity.


sorbits:
Merged as f3cd0b079, thanks!

Was the issue only a warning, or did you see layout problems (I do understand that the layout is ambiguous though)?

rdwampler:
> Was the issue only a warning, or did you see layout problems (I do understand that the layout is ambiguous though)?

A single warning was logged in the Console and I did not see any visible layout issues.

PR #1383, opened 2017-04-19
RFC: Add support for Finder Tags

rdwampler:
Requires building with 10.9 SDK

I am planning on adding the "tagging" popover like in Finder, but wanted to get any comments on the overall approach/look before continuing.

### DIFF
--- a/Frameworks/OakFileBrowser/src/FSOutlineViewDelegate.mm
+++ b/Frameworks/OakFileBrowser/src/FSOutlineViewDelegate.mm
@@ -7,6 +7,7 @@
 #import <OakAppKit/NSColor Additions.h>
 #import <OakAppKit/NSImage Additions.h>
 #import <OakAppKit/OakAppKit.h>
+#import <OakAppKit/OakFinderTag.h>
 #import <OakAppKit/OakUIConstructionFunctions.h>
 #import <ns/ns.h>
 #import <io/path.h>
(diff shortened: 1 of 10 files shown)


posguy99:
Can I just say, YES!  I want.  (just a random user comment here ^_^ )

jeremydouglass:
Excited about Finder tag support.

jakecast:
I'd also like to see this added in!

sorbits:
@rdwampler Is there anything preventing this from being merged?

I held off doing a review since it was marked RFC and not a small patch.

rdwampler:
> @rdwampler Is there anything preventing this from being merged?

At the time, the only barrier was the 10.9 sdk requirement. It's not feature complete (i.e., adding custom tags via a popover) and some visual tweaks might be necessary. I would also like to make the `OakLabelSwatchView`  base on an `NSImageView`, but I think the current state should be OK as a start.

I've been busy lately, but I hope to have some time in a few weeks to pick this back up.

rdwampler:
@sorbits This lacks the ability of adding custom tags, but do you think it's in a state to merged? Let me know what is need and I try to jump back onto this. Thanks!

Frameworks/OakFileBrowser/src/FSOutlineViewDelegate.mm
```diff
@@ -219,17 +277,17 @@ - (void)updateConstraints
...
 
-- (void)setLabelIndex:(NSInteger)newLabelIndex
+- (void)setFinderTags:(NSArray<OakFinderTag*>*)newFinderTags
 {
-	if(self.labelIndex == newLabelIndex)
+	if([self.finderTags isEqual:newFinderTags])
 		return;
 
-	if(newLabelIndex == 0)
+	if(!newFinderTags && [newFinderTags count] == 0)
```
sorbits: This should probably be `if(!newFinderTags || [newFinderTags count] == 0)` or possibly just `if(![newFinderTags count])` (relying on sending `count` to `nil` returning `nil`).

kfjelsted:
If tags support is being added please make sure that the entire file dialog including tags is still supported for accessibility via the Voiceover screen reader!
Thanks so much!
-Kevin

Frameworks/OakFileBrowser/src/FSOutlineViewDelegate.mm
rdwampler: Thanks for catching this!

sorbits:
> If tags support is being added please make sure that the entire file dialog including tags is still supported for accessibility via the Voiceover screen reader!

This pull request changes TextMate’s current support for assigning a single label to a file, into assigning multiple tags to the file.

This is done via the action menu in the file browser, however, it is currently not accessibility compliant, but the previous label system wasn’t either, and the save dialog is unaffected by this pull request, so this pull request should not break anything per se.

As for making the tag system accessible, I think this is something we can handle and I appreciate that you are calling attention to this.

sorbits:
@rdwampler Merged the PR and pushed a few additional commits, thanks for all the work!

Some thoughts:

- It seems there is no way for us to know what colors are associated with tags, right? This is especially a problem if I make a custom tag in Finder, give it a color, and then assign it to a file in TextMate. ~~~The result is that the tag is assigned but with no color.~~~ **Update:** Now it seems to work.

- [ ] The system deals with multi-select slightly differently: If multiple items are selected, and selecting a tag that is not set for all the items, then the action is interpreted as “set tag”, where in TextMate it is “remove tag” when 2+ items have the tag set.

- [ ] We can probably move tag reading/writing to `FSItem` to ensure proper KVC compliance and avoid having to pass around an `std::string`/`NSData` with tag info. If `FSItem` loads tags lazily, it will only do this for the items visible in the file browser (due to view-based table views and bindings), so the performance hit should be negligible from not pre-loading tags in a background thread.

- [ ] The `OFBFinderTagsChooser` should probably use subviews for the color swatches. This would simplify mouse actions and also make it easier to enable accessibility. Could possibly use the `OakRolloverButton` although that would require an `NSImage` subclass to draw the color circles.

- [ ] The list of “favorite tags” should ideally be fetched each time the menu is opened rather than only once per session.

- [ ] I find the “favorite” prefix somewhat confusing. I see the Finder user defaults key also refers to them as favorite tags, but anything besides the `OFBFinderTagsChooser` does not care about favorites, right? And for `OFBFinderTagsChooser` it is the list of tags that must always be shown, in addition to the tag(s) set for the selected item(s). Or did I miss something?

- [ ] I would prefer to see the label/color stuff abstracted by `OakFinderTag`. Right now we ask if tags have labels, the label number, and then foreground/background color for these labels. Could we get away with nullable `NSColor*` foreground/background properties?

- [ ] Related to the above, I would like to see using `OakFinderTag` directly, instead of its display name or label. See e.g. 88757f25bfcb29495f4b4d860fb0155c8d85691f where we write the tag names to disk, and then read back the tags to get the proper `OakFinderTag` for updating the `finderTags` property. This commit was me being lazy, I should have refactored the method, but `OFBFinderTagsChooser` also delivers the selected tag using its display name, so this is a general refactoring item.

- [ ] I think it would make sense to have separate `addFinderTagAction` and `removeFinderTagAction` properties on `OFBFinderTagsChooser` rather than have the action method need to access both `chosenTag` and `removeChosenTag` to decide what to do.

- [ ] Should remove all (legacy) code related to labels.

I made the above items check boxes as I believe most of it is actionable. I’ll check them off if I do some of it myself (I’ll likely look at some of the refactoring stuff now).

sorbits:
@rdwampler I just remembered about the simplified `NSImage` constructor that takes a block as the draw handler (10.8), so I made the control use `OakRolloverButton` in 1002c868c3ed04a62b5c1e535d8d3784fca92849 which gives us Voice Over support and eliminates the need for implementing mouse up/down.

I can’t think of a simple way to update the label based on mouse position without introducing notifications posted by `OakRolloverButton`, so that is probably the route I’ll take, as it would allow to remove all the remaining code related to tracking rects and mouse movements.

/cc @kfjelsted

rdwampler:
@sorbits Thanks for working on this!

> I can’t think of a simple way to update the label based on mouse position without introducing notifications posted by OakRolloverButton, so that is probably the route I’ll take, as it would allow to remove all the remaining code related to tracking rects and mouse movements.

I was going down this approach too.

PR #1385, opened 2017-04-26
Add "Copy As Pathname" to file browser's context menu

rdwampler:
This is implemented as an alternate menu item to "Copy" like in Finder.

I discovered this recently in Finder and I find very handy when interacting with the terminal. I wanted to implemented in the main menu but, of course, ⌥⌘C is already in use.

### DIFF
--- a/Frameworks/OakFileBrowser/src/OakFileBrowser.mm
+++ b/Frameworks/OakFileBrowser/src/OakFileBrowser.mm
@@ -758,6 +758,20 @@ static bool is_binary (std::string const& path)
 	[self writeSelectionToPasteboard:[NSPasteboard generalPasteboard] types:nil];
 }
 
+- (void)copyAsPathname:(id)sender
+{
+	NSMutableArray* pathnames = [NSMutableArray array];
+	for(NSURL* url in self.selectedURLs)
+	{
+		if([url isFileURL])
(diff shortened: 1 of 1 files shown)


sorbits:
> I discovered this recently in Finder and I find very handy when interacting with the terminal

If I use the regular copy with selected items in the file browser and paste in either iTerm or Terminal, it inserts the full path.

In what situation does regular copy not work?

rdwampler:
You're right. But it's consistent between apps (e.g., with the regular copy command, TextMate pastes just the basename, Mail as a file attachment, etc.) One thing to note though is it copies the path as a plain text string and will not perform any escaping if pasted into Terminal unlike the regular copy. This behavior is consistent with command in Finder.

sorbits:
OK, merged as 08f41a37d :)

As for the key equivalent, it sort of seems like <kbd>⌥⌘C</kbd> should be context dependent, so at least with focus in file browser, it would be _Copy as Pathname_, unfortunately once something is in the main menu, it’s hard to give it a context sensitive behavior, unless we reuse the same menu item, and these two (_Ignore Case_ and _Copy as Pathname_) are rather different.

The only reason for having _Ignore Case_ in the menu is for discoverability, as a lot of people have asked for key equivalents to toggle options in the find dialog.

PR #1390, opened 2017-06-23
Fixes for newer versions of Clang

rdwampler:
This gets TextMate to compile with newer versions of Clang with c++1z; but it crashes at runtime. The crashes look like lifetime issues and only occur when compiling with `-std=c++1z` and not `-std=c++14`. I will send the crash reports via the mailing list.


jakecast:
ah, thats a handy fix.. `-std=c++14` got mine to compile under 10.13 w/ Xcode 9

PR #1391, opened 2017-08-24
Enable ThinLTO

jgavris:
Enables fast link-time optimization available since clang 3.9. 

https://clang.llvm.org/docs/ThinLTO.html

### DIFF
--- a/target
+++ b/target
@@ -14,15 +14,19 @@ FLAGS += -DNDEBUG -Os
 # LINK   = OakDebug
 
 CXX_FLAGS    += -fvisibility=hidden -std=c++1z -stdlib=libc++
+CXX_FLAGS    += -flto=thin
 OBJC_FLAGS   += -fvisibility=hidden -fobjc-arc -std=c99 -fobjc-abi-version=3
+OBJC_FLAGS   += -flto=thin
 OBJCXX_FLAGS += -fvisibility=hidden -std=c++1z -stdlib=libc++ -fobjc-abi-version=3
 OBJCXX_FLAGS += -fobjc-arc -fobjc-call-cxx-cdtors
+OBJCXX_FLAGS += -flto=thin
 LIBS         += c++
 
 LN_FLAGS += -m64 -mmacosx-version-min=$APP_MIN_OS
 LN_FLAGS += -fvisibility=hidden -stdlib=libc++
 LN_FLAGS += -Wl,-dead_strip
 LN_FLAGS += -Wl,-dead_strip_dylibs
+LN_FLAGS += -flto=thin
 
 FLAGS     += -I"$libressl_prefix/include"
 CXX_FLAGS += -I"$capnp_prefix/include"


sorbits:
This is interesting, thanks — unfortunately the current cost (in build time) is too high.

With ThinLTO it took ~30s to do a new build after touching just one source, compared to ~2s w/o ThinLTO.

Though I’ll leave the PR open as a reminder of this feature.

jgavris:
Perhaps there's something else going on here...this was a quick and dirty effort. ThinLTO is supposed to be on-par with a regular -O2 build 

http://blog.llvm.org/2016/06/thinlto-scalable-and-incremental-lto.html

>The results are for an end-to-end build of clang (ninja clang) from a clean build directory, so it includes all the compile steps and links of intermediate binaries such as llvm-tblgen and clang-tblgen. Release build shows how ThinLTO build time is very comparable to a non-LTO build.

rdwampler:
I believe the above timings is for a full build. We probably have to look at "Incremental ThinLTO Support", AFAICT. Although building just the release binary under THinLTO may be the way to go and keep the current settings for development.

jgavris:
Ninja's 'way' is to avoid conditional compilation rules though, right? I didn't see infra for that and as ninja is not my most familiar build system.

jgavris:
AFAIK `-flto=thin` is thin / incremental LTO. How do you build the release binary?

sorbits:
> AFAIK -flto=thin is thin / incremental LTO. How do you build the release binary?

If you do a build then run `touch Applications/TextMate/src/main.mm` followed by `ninja -vn TextMate` then you will see the 3 lines that `ninja` will execute (recompile `main.mm`, link the `TextMate` executable, and lastly sign the application bundle).

On my MBP this takes ~17 seconds (my time above was for touching a file that required two executable to be linked).

jgavris:
Great, thanks for sharing. There is no separate 'release configuration' which is used for building to distribute?

rdwampler:
> Ninja's 'way' is to avoid conditional compilation rules though, right?

Right. The current build system (target files) doesn't support conditional compilation neither. To build a release binary release build you would have to change the main target by hand each time (like  a DEBUG build you need to uncomment the lines `# FLAGS += -D_GLIBCXX_DEBUG` and `# LINK   = OakDebug`, etc.)

To enable incremental LTO, add the following line `LN_FLAGS += -Wl,-cache_path_lto,/path/to/cache` a incremental build goes from ~17 sec to ~3 secs.

jgavris:
Ahh! Very nice find @rdwampler!

jgavris:
I'm seeing < 1s for a touch / build with the cache path set.

sorbits:
Merged as d1289e0f6dba953a412b8271f055e944bbb68a8d (with slight modifications, including setting up the cache folder pointed out by @rdwampler).

PR #1393, opened 2017-10-02
Update compiler flags for High Sierra (10.13)

rdwampler:


sorbits:
Merged as 239d5bc19...234e64dcf.

Had been running into this myself, although I actually use Xcode 7 to build TextMate because Xcode 8 has introduced an issue where the file browser is prone to crash (and I have not been able to locate the problem, as the crash is in `NSOutlineView` and seemingly triggered by some internal garbage collection, so somewhat random and without any meaningful TextMate code on the stack).

Thanks to your patch I can now test if Xcode 9 fixes this :)

PR #1395, opened 2017-10-05
Touch bar extensions

rdwampler:
This PR begins to extend the initial touch bar support to OakFileBrowser and OakTextView. It also makes each touch bar completely customizable.

For OakFileBrowser, the Finder makes a lot of items available to the touch bar despite being accessible via keyboard shortcuts or the toolbar so I plan to mimic this by adding more items like "Move to Trash", "Quick Look", etc..

@jakecast I would appreciate any feedback. I don't have an opinion on the default set of items, except for setting and navigating bookmarks, which I think should always be the default. These are customizable though so they can be removed.


sorbits:
It’s worth mentioning that the semantic classes assigned to bundle items were originally done to allow adaptable toolbars. For example all “run” actions have a semantic class prefix of `process.run`, but I never implemented the query language I had in mind for this, but it does seems like this could be repurposed for the touch bar.

rdwampler:
I thought about supporting bundle items too. I am willing to prototype something. Do you mind sharing your ideas on how the querying should work?

dantiel:
i think supporting this in a programmable way be definitely a sugarish feature in a customisable editor, hence being connected to one of text mates keyfeatures, especially as duet for example supports an on screen touchbar as well, which may fit an important gap in using native mac apps with touch devices (gestures are not as necessary as the user is anyways still more apt to return to the keyboard in this kinda workflow).

jakecast:
I really, really dig it! Depending on how the bundle items are query'd and supportable, it might be worth trying to add support for the 'Customize Touch Bar...' view that is starting to become the default in many macOS apps. 
That said I'm not sure how well that control would work if we wanted to allow customization on a per-language settings. (e.g. this set of items for C++, this set for Swift, etc.)

dantiel:
i think a good approach would be by extending the interface in a way that allows bundle execution or macro programming, e.g. in the shell `mate` cmd to have a way to properly address a bundle command identifier via `process.run`. instead of linking from the outside back to an outside, text mate could be even more generous in exposing parts of its internal structures meaning that events are apt to trigger from different sources by providing a context through some nebulous way of controllable parameters. that seems only one part of it. the other is, that yea a toolbar is a nebulous extension for being mostly text focused, and language specific toolbars start to look like an IDE feature. furthermore, to have all control over touchbar extensions one would at least have to write a plugin e.g. for each component instead of a bundle only.

skurfer:
> I don't have an opinion on the default set of items, except for setting and navigating bookmarks, which I think should always be the default.

If all we had were buttons for setting/navigating bookmarks, that would be a huge step. I almost never use them because the keystrokes are so awkward.

mxmerz:
I don't have any knowledge of TextMate's internals and can't help with an implementation, but I would really appreciate improved touch bar support. Has there been any progress on this since October 2017?

rdwampler:
@sorbits I recently moved to doing most of my work on the new MacBooks, so I would like to pick this PR back up. I rebased the PR on top of master and dropped the changes to the File Browser for now. The changes are now limited to only adding the ability to add/remove (and navigating) bookmarks and adding accessibility support. 

Is there anything blocking this from getting upstream?

sorbits:
> Is there anything blocking this from getting upstream?

I have merged as 1ef78b3af...b71efb006, thanks!

> Do you mind sharing your ideas on how the querying should work?

Unfortunately I can’t find my notes about this, but if you open _Select Bundle Item_ and enable all scopes and switch to semantic class you can do a simple query.

For example `style.strong` will show you all commands that turn text bold, so by this simple query, we can have a universal Bold button in the touch bar.

Likewise `process.build` shows all build commands (as mentioned in my old comment above).

For shortcuts (key bindings) this is where we need an actual query language, since one key may be used for different types of items, depending on what’s available, presumably simple lazy chaining would be sufficient for must cases, e.g.:

    ⌘B = style.strong || process.build || process.run

By giving the query used for shortcuts a scope selector, it would be possible to alter the above query based on whether or not there is a selection, if we are in a specific file type, etc.

But the touch bar can show all matches and with different icons, so maybe simple queries would actually be enough here.

PR #1396, opened 2017-10-05
Explicitly close OakChoiceMenu

rdwampler:
I am on High Sierra and I am noticing that OakChoiceMenu no longer closes unless we explicitly call `close` on the window. I don't know how this will behave on 10.12 or older version. Was there a reason we never called `close` ourselves before?


sorbits:
> Was there a reason we never called close ourselves before?

Not per se, but `self.choiceMenu = nil;` should release the choice menu → cleanup → release window → close window.

So I fear this might reveal a memory leak.

Can you redo this patch so that we instead call `[self.window close]` in `OakChoiceMenu`’s `-
 dealloc`? That way, the API remains the same, and if it doesn’t work, there most likely is a leak.

PR #1398, opened 2017-10-13
gen_credits: create the database file and enclosing directory as necessary

chucklebot:
and handle API rate-limiting as well, which can easily happen to eager hobbyists

### DIFF
--- a/bin/gen_credits.rb
+++ b/bin/gen_credits.rb
@@ -6,6 +6,7 @@
 $KCODE = 'U' if RUBY_VERSION < "1.9"
 
 require 'digest/md5'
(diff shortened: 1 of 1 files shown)


ryandesign:
> and handle API rate-limiting as well, which can easily happen to eager hobbyists

Or anyone. If a new version of textmate is committed to MacPorts, for example, that will kick off automated builds on our build farm on 6 different macOS versions simultaneously. Perhaps that's enough to trip the rate limit.

sorbits:
Merged as 9f2eb7622...108b99149, thanks!

Split into two commits since it’s two different changes.

PR #1399, opened 2017-10-18
Re-enable building with c++1z

rdwampler:
c++17 now guarantees eliding copys when initializing objects from prvalues. This causes problems when using the GNU extension `?:` operator and when the prvalue is used in method parameters. To force a copy, we cast them to rvalues.

This feels very hacky, but I don't know if there is anything else we can do.


rdwampler:
Closing this since, the issue around `?:` is fixed in clang repo and there is a patch submitted to resolve the second problem with setting properties via dot syntax.

PR #1400, opened 2017-11-11
allow view plain text files without a file extension in quick look

pravdomil:
such as README, LICENCE, makefile etc..

### DIFF
--- a/Applications/QuickLookGenerator/resources/Info.plist
+++ b/Applications/QuickLookGenerator/resources/Info.plist
@@ -44,6 +44,7 @@
 				<string>public.source-code</string>
 				<string>public.plain-text</string>
 				<string>public.text</string>
+				<string>public.data</string>
 			</array>
 		</dict>
 	</array>


sorbits:
Merged as fd827fb81, thanks!

I reworded the commit message to give a bit more detail about the change.

jmeosbn:
Hi there, this seems also enable quicklook for file types with an extension (not known to text mate?) such as iTunes libraries, dmg files, etc..

sorbits:
> this also enables quicklook for file types with an extension (not known to text mate?), such as iTunes libraries, dmg files, etc.

I assume you are referring to thumbnail creation?

A few users have brought this up, and I see no way to conditionally opt out of thumbnail creation, so I am afraid we’ll have to revert this commit.

@pravdomil Any comments?

sorbits:
> quicklook extension itself could check if the public.data file is binary and if so don't render quicklook

The problem is with the thumbnail generation, the preview itself I think doesn’t matter, as people generally do not quicklook binary files.

But if you open a folder (in Finder) with disk images etc. and use Icon View you will get “thumbnails” from the QL generator with the `public.data` commit.

I tried to make the code not render these thumbnails, but it seems that if something is registered for the file type, a thumbnail *must* be created, so if TextMate’s QL plug-in didn’t generate anything, it just showed a generic text file thumbnail (rather than the proper disk image icon or similar)

sorbits:
> I tried to make the code not render these thumbnails, but it seems that if something is registered for the file type, a thumbnail must be created, so if TextMate’s QL plug-in didn’t generate anything, it just showed a generic text file thumbnail (rather than the proper disk image icon or similar)

The only partial solution I found was to set the minimum size to something huge, but that would disable all thumbnail generation.

pravdomil:
> The only partial solution I found was to set the minimum size to something huge, but that would disable all thumbnail generation.

if you set file limit to eg. 1MB and you set `Icon View` in `Finder` what thumbnail you will get for eg. 10MB `dmg` file?

sorbits:
There is only `QLThumbnailMinimumSize` to set the minimum point size of 
thumbnails for which the generator should be called.

So e.g. if you are in Icon Mode but icons are 48×48 and 
`QLThumbnailMinimumSize` is set to 128 then you will get the default 
document icons.

But if the generator itself just refuse to generate icons (e.g. `return 
-1` if the file type is `public.data` with an extension) then the system 
will *not* show the default document icon, but instead it seems to 
fallback on the standard text file thumbnail generator.

jjeising:
> The problem is with the thumbnail generation, the preview itself I think doesn’t matter, as people generally do not quicklook binary files.

I would personally rather disable preview for `public.data` if there is no other way. Getting filename and size in Quick Look is much more helpful than TextMate rendering text from .zip, .dmg and other very common files.

There should be at least a binary check (if possible), this would also help to detect slightly wrong file extensions (recently typed `xlxs` instead of `xlsx` and wondered for minutes about the TextMate preview).

danmackinlay:
FWIW I've just had an agonising 3 weeks of Finder crashing which I eventually traced to TextMate's excessively permissive `.qlgenerator`.

Since I don't have acccess to the issue tracker I documented the problem here, and send supporting crash logs to the support email address on the macromates site.
https://livingthing.danmackinlay.name/text_editors.html#tm-ql-crash

I don't know, of course, if this crash is purely due to the Textmate .qlgenerator, but even if this problem is due to some weird interaction with third party software, it's a good indicator that the textmate qlgenerator should be much more conservative

PR #1403, opened 2018-02-07
Add imported and exported type declarations for D

jacob-carlborg:
This will make the system properly recognize D source files. I'm not sure if this is correct. I did the same as for C.


sorbits:
Merged as e3ded6e2f9ede6b9f5fb7f7f3022b621355a2df4.

Change icon to `Blank` since there is currently no `D` icon.

jacob-carlborg:
Thanks.

PR #1404, opened 2018-04-04
Always uppercase the encoding charset

rdwampler:
Apple writes the encoding charset (e.g., com.apple.TextEncoding) in
all lowercase (at least since macOS 10.9); however, TextMate assumes that the
encoding will always be uppercase. So in order to avoid any issues with case
mismatches, when setting the charset, uppercase the given string.

See http://lists.macromates.com/textmate/2018-April/040602.html

### DIFF
--- a/Frameworks/file/src/encoding.h
+++ b/Frameworks/file/src/encoding.h
@@ -3,6 +3,7 @@
 
 #include "bytes.h"
 #include <oak/oak.h>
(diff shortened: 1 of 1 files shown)


sorbits:
Merged as 1b33f6e36daefc48d9ba86822738c71edd22cba0, thanks for tracking this down and fixing it!

PR #1405, opened 2018-05-25
Switch to TextMate.app when launching from mate

jgavris:
This has been annoying me for a bit as I often have TextMate buried under a bunch of other applications. I've had to `mate` and then command + tab to it before I can start typing. Curious what the motivation was for this originally?

### DIFF
--- a/Applications/mate/src/mate.cc
+++ b/Applications/mate/src/mate.cc
@@ -85,7 +85,7 @@ static void launch_app (bool disableUntitled)
 
 	cf::array_t args(disableUntitled ? std::vector<std::string>{ "-disableNewDocumentAtStartup", "1" } : std::vector<std::string>{ });
 
-	struct LSApplicationParameters const appParams = { 0, kLSLaunchDontAddToRecents|kLSLaunchDontSwitch|kLSLaunchAndDisplayErrors, &appFSRef, nullptr, nullptr, args, nullptr };
+	struct LSApplicationParameters const appParams = { 0, kLSLaunchDontAddToRecents|kLSLaunchAndDisplayErrors, &appFSRef, nullptr, nullptr, args, nullptr };
 	OSStatus err = LSOpenApplication(&appParams, nullptr);
 	if(err != noErr)
 	{


sorbits:
This [`- bringToFront`](https://github.com/textmate/textmate/blob/master/Frameworks/DocumentWindow/src/DocumentWindowController.mm#L2691) function gets called when using `mate` to open one or more documents, which should ensure the application gets activated.

The reason that `mate` does *not* ask to get TextMate activated is because at that time, no new windows have been created, and ideally when calling `mate` from a terminal we only want the *new* document window to obscure the terminal (being brought to front), and have any existing TextMate windows stay in the background. But if we were to activate TextMate, at least one window would go to front *before* opening the new window.

Given your PR it sounds like something is not working as it should on your system, I’m closing this PR because I do not think your commit is necessary, but I would of course like to know why `mate`/TextMate is not working as intended on your system.

jgavris:
Hm, that makes sense. My typical workflow is writing some code in Xcode, going to my terminal, staging a commit, and then writing the commit message (or interactive rebase) in TextMate (via `mate -w`). It seems that when I have Xcode in the foreground, TextMate does not activate until I take focus away from the terminal (by clicking into Xcode). I focus Xcode and then TextMate jumps into the foreground.

sorbits:
> TextMate does not activate until I take focus away from the terminal (by clicking into Xcode). I focus Xcode and then TextMate jumps into the foreground

This sounds rather strange, so basically TextMate activates, but with a delay?

I can’t really think of why that would be. Is this 100% reproducible? Try with just `mate -w` and maybe other applications than Xcode, see if you find a pattern.

dragonsinth:
Sad this PR was closed... this has been bother me for over a year now.  I have my EDITOR env var set to `mate -w` and it used to work beautifully, but since some update a while back, I have to always tab over to the already-running TextMate app if I don't want to wait forever for it to pop up.  Really a drag for the many frequent git commands I run that need an editor.

koenpunt:
I actually think this is not TextMate specific; I currently have it when I select "Reveal in Finder" from any application; Finder opens only after some seconds, or I even have to tab out of the current application, and activate Finder before it actually opens a new window to reveal the selected item.

sorbits:
> Sad this PR was closed

We need to know exactly what the problem is before applying a fix.

If you want to help track it down then upgrade to latest build (⌥-click Check Now in Preferences) and see [this message](https://lists.macromates.com/textmate/2019-August/041086.html) and [this one](https://lists.macromates.com/textmate/2019-July/041038.html) for how to extract the extra debug log.

dragonsinth:
> I actually think this is not TextMate specific; I currently have it when I select "Reveal in Finder" from any application; Finder opens only after some seconds, or I even have to tab out of the current application, and activate Finder before it actually opens a new window to reveal the selected item.

I dunno, if I `open Makefile`, it comes up in TextEdit immediately (but of course, the command line won't block on it).  Whereas if I `mate -w Makefile` it spins for me until I wait like 5 seconds or else tab over to TextMate.app.

> If you want to help track it down then upgrade to latest build (⌥-click Check Now in Preferences) and see [this message](https://lists.macromates.com/textmate/2019-August/041086.html) and [this one](https://lists.macromates.com/textmate/2019-July/041038.html) for how to extract the extra debug log.

Now that's interesting!  I didn't know about test builds, but I just installed  2.0-rc.30... and it seems to have fixed the problem entirely!  I guess this issues is rightfully closed, we just need a new release. :)

jaysoffian:
> If you want to help track it down then upgrade to latest build (⌥-click Check Now in Preferences) and see this message and this one for how to extract the extra debug log.

@sorbits: https://gist.github.com/jaysoffian/86a3316892cbca10ff41be15b8691349

sorbits:
> @sorbits: https://gist.github.com/jaysoffian/86a3316892cbca10ff41be15b8691349

Am I missing something, or is this the information from rc.29 that I already commented on (adding more debug output to rc.30 and giving instructions about how you should be able to run `sample` to get a stack trace when the issue appears).

jgavris:
For anyone else who finds there way here, I believe @sorbits has resolved this in rc.31 / https://github.com/textmate/textmate/commit/7116dd777f07a65a675fe8944bc2fc2f05cada72.

PR #1406, opened 2018-05-30
Add ellipsis to update check menus

antons:
This matches most other apps with a software update mechanism.

This patch is released in public domain.


sorbits:
Ellipsis indicates that the menu item doesn’t perform any action but (normally) opens a dialog from where the user can perform the action (or task) indicated by the menu item.

In this case, the menu item is “Check for Updates” and that is being done immediately after the item has been selected (with no intervening dialog).

This is contrived, but imagine a user on a non-free internet connection, they may think that selecting an item named “Check for Updates…” would not actually use any bandwidth, because the item indicates that the user will be prompted with options before performing the check.

Ellipsis would make sense if the item was named “Update to Latest Version…” because that action would require user interaction before going ahead.

I didn’t do a survey myself of apps, so I do not know if app developers disagree with this, but I do seem to recall that back when Apple had a “Check for Software Updates” item in the system () menu it was without ellipsis, and looking in iTunes, I see this:

image

So I do believe that TextMate’s lack of ellipsis is correct.

But I am leaving your PR open because I am open to input.

antons:
Thank you for the considered response, @sorbits.

Quoting [HIG](https://developer.apple.com/macos/human-interface-guidelines/menus/menu-anatomy/):

> Use an ellipsis whenever choosing a menu item requires additional input from the user. The ellipsis character (…) means a dialog or separate window will open and prompt the user for more information or to make a choice.

You’re correct, most often the menus with ellipsis don’t actually perform an action or task. But HIG doesn’t mention the action or task specifically, only the expectation of further input.

By most apps I meant those that use Sparkle. Sparkle opens a window with release notes and prompts the user to make a choice: Skip Update, Remind Me Later, or Install Update. This matches what HIG prescribes. Sparkle [recommends](https://sparkle-project.org/documentation/) naming the menu item as “Check for Updates…”.

TextMate opens a very similar window, and prompts the user to make a choice: Later, or Download & Install. I think that including the ellipsis in TextMate would still fit with what HIG prescribes and match what’s done by most apps.

antons:
P. S. IIRC, you’re right about the old “Check for Software Updates” item, and yes, iTunes names its button without ellipsis. But on the other side we have Transmit, Coda, 1Password, Espresso, OmniFocus, OnmiPlan, BBEdit, Tower, just to name a few.

infininight:
Seems the question here is how you interpret the menu item. "Check for Updates" as the title says is performed immediately when you select the item without further input or options from the user. This would suggest it should not have an ellipsis. However the response from the command may present an option if there is a new version available which does require a choice from the user. For this half probably an ellipsis is proper.

I think because the main intent of the command is the checking half the ellipsis isn't proper, but it's not a clear resolution.

koenpunt:
To create clarity maybe the option should be "check for updates..." (including ellipsis), which when selected opens a new dialog, which displays details about the current version, and also a "check for update" button. 
But creating this just to satisfy using an ellipsis in a menu item is probably overkill.

sorbits:
I’ve decided not to make this change. Besides the reasoning stated previously, if we were to make this change, I’d argue that for consistency the “Check Now” button in Preferences → Software Update should also have ellipsis, but that seems even more wrong to me.

The PR did however make me realize that the title should be changed from plural to singular, as we are only checking for one update, I made this change in b577c580c41f3a0fac701ac9bace8f77ae89f73f.

PR #1407, opened 2018-07-08
Add excludeSCMDeleted to hide deleted files

ThatsJustCheesy:
Add a new setting that prevents deleted files from appearing in the file browser while still uncommitted. There [seems](https://stackoverflow.com/questions/17590302/textmate-2-alpha-showing-references-for-deleted-files) to be some [demand](https://stackoverflow.com/questions/50428259/how-can-i-keep-deleted-files-from-showing-up-in-textmate-2-file-browser-sidebar) for this, and it's just a nice option to have.

Also fixed some apparent typos in FSSCMDataSource.

EDIT: Also, I hereby dedicate this patch to the public domain.


sorbits:
I was hesitant to merge this because without the ghost files it is not possible to commit a rename or delete from within TextMate, and as one of the two posts you linked to show, the user who wanted them gone was unaware that their disk and git index were out of sync, so in that case, the ghosts was a useful cue that something was wrong.

But given the small footprint of having the option I merged the commits (with a minor change) as 9240fd008...096801b8a.

PR #1409, opened 2018-08-19
Dark mode support for tab bar

rdwampler:
Light Mode:

screen shot 2018-08-19 at 6 53 39 pm

Dark Mode:

screen shot 2018-08-19 at 6 52 53 pm


Frameworks/OakTabBarView/src/OakTabItemView.mm
```diff
@@ -151,8 +200,35 @@ - (id)init
...
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunguarded-availability"
+			NSAppearanceName appearanceName = [[NSApplication sharedApplication].effectiveAppearance bestMatchFromAppearancesWithNames:@[ NSAppearanceNameAqua, NSAppearanceNameDarkAqua]];
+			if([appearanceName isEqualToString:NSAppearanceNameDarkAqua])
+				_images = [self mojaveDarkImages];
+			else
+				_images = [self mojaveLightImages];
+#pragma clang diagnostic pop
+
```
rdwampler: This section should really be enclosed in an `@available` check, but I was trying to make sure this would still compile with Xcode 7.

sorbits:
Thanks, I appreciate trying to preserve Xcode 7 compatibility, although the pragma results in an `unknown warning group` warning, so I removed this part, but added the necessary API definitions to [`sdk-compat.h`](https://github.com/textmate/textmate/blob/master/Shared/include/oak/sdk-compat.h).

Don’t think Xcode 7 can use an SDK beyond 10.11 because the SDKs use some of the new Objective-C features (like class properties).

I will still be building release versions of TextMate with the 10.11 SDK (until I figure out a workaround for the crashing file browser when using Xcode 8+), this means Dark Mode won’t be available to users unless we set `NSRequiresAquaSystemAppearance` to `NO` in `Info.plist`, but I don’t yet know how feasible it is to set this, as I am not on the Mojave betas, e.g. the status bars and file browser header should also be updated to support Dark Mode.

I didn’t merge 15712ab7ca3b9186a963caa2c1b2a63514708cbf because if we do that change, we should also update the colors to have more contrast between selected and non-selected tabs.

rdwampler:
`NSRequiresAquaSystemAppearance` should be set to `YES` for the reasons you mentioned, in addition to others (E.g., `OakSyntaxFormatter` needs to be updated to include a dark theme). 

Another issue when building with the 10.14 SDK is that the icon images in the file browser and `FileChooser` are not displayed. I haven't track down why yet. Maybe because they're being drawn off screen? Using of the NSImage block base API may resolve it, but idk.

PR #1412, opened 2018-10-07
An attempt for better dark mode support

rdwampler:
Using NSVisualEffectView allows us to support "Desktop Tinting in Dark Mode". There's still a lot work to be done concerning the dividers.

screen shot 2018-10-06 at 11 42 12 pm
screen shot 2018-10-06 at 11 52 04 pm


sorbits:
Merged as f6fa27391...ef032c78a, thanks!

The API for the status bar effect view is available on 10.10 and later, so I enabled the effect there, but had to make the containing view layer backed, which is also mentioned in the documentation, but I vaguely recall something about all views now being layer backed, which is probably why you did not see an issue on 10.14.

PR #1413, opened 2018-10-07
Added documentation

timgiroux:


sorbits:
Appreciate the sentiment, but even with these comments, I still think one has to read the code to understand what goes on.

If the boolean argument is confusing, changing the API would be better.

PR #1415, opened 2018-10-09
Don't update our background style unless it changes

rdwampler:
Also, always remove our background view before attempting to add another one.

@sorbits, I'm not sure if this could be a factor for the crash in #1414, but this obviously wasn't doing the appropriate thing before.

/cc @ThatsJustCheesy


Frameworks/OakAppKit/src/OakUIConstructionFunctions.mm
```diff
@@ -202,13 +202,22 @@ - (NSSize)intrinsicContentSize
...
 }
 
 - (void)updateBackgroundStyle
 {
+	if(!_visualEffectBackgroundView)
+	{
+		[_visualEffectBackgroundView removeFromSuperview];
+		_visualEffectBackgroundView = nil;
+	}
+
```
sorbits: This doesn’t look right to me: Only remove the view when there is no view?

PR #1419, opened 2018-10-29
Change Select Previous/Next Tab key equivalents to system preferred ones

rdwampler:
⌃⇧⇥ and ⌃⇥ are the preferred system wide shortcuts for changing tabs. In order to avoid a disruptive change, the previous shortcuts are still supported (in fact, it appears that all apps using the system tabbing mechanism support the previous shortcuts as well).


sorbits:
Eventually we may also have to reassign <kbd>⌘T</kbd> to _New Tab_ but that’ll be a hard pill to swallow, need a similarly easy key to reach for _Open Quickly_.

PR #1422, opened 2018-11-02
Rw/change theme based on appearance

rdwampler:


sorbits:
Merged as 663a9a638...5b686f66e.

I added default themes so it is enough to set `changeThemeBasedOnAppearance`.

I am thinking a way to handle the UI is something like this:

Add checkmark item to View → Themes that sets whether or not the theme should be “dynamic”.

When the user changes theme via the View → Themes menu, we store it under the appropriate defaults key (Dark or Universal).

That way, when the global appearance is changed to e.g. light, TextMate will change to the last light theme that the user picked, and likewise for dark.

One thing to remember is that during startup, we should pick the correct theme, as appearance may have changed while TextMate was not running.

rdwampler:
@sorbits I'll work on your suggestion, thanks!

PR #1425, opened 2018-11-14
Use a divider with a flat appearance

rdwampler:
This seems consistent with the divider style in the header view of Finder in the list view and the overall system flat look since 10.10.


rdwampler:
Agree. I had considered using NSBox too, but thought that NSImageView was simpler. :)

PR #1426, opened 2018-11-16
Textmate 19 and before....Python output

tatecapone:
A simple python file with a print with relative and absolute path does not show the last folder. The one after Boring Stuff/`07-Files`
captura de pantalla 2018-11-16 a la s 16 12 34

The same result in the terminal.
captura de pantalla 2018-11-16 a la s 16 13 49


sorbits:
If this is a bug report please see https://macromates.com/support

tatecapone:
ok. Thanks you.

PR #1427, opened 2018-12-02
Migrate OakTextView to new (10.10+) Accessibility API

dusek:
More description in commit messages.

Tested with VoiceOver and Zoom on macOS 10.14.1 (18B75) to make sure it works as expected after the migration.

This makes progress on textmate/bugs#13. For time reasons, I left the `OakAccessibleLink` in OakTextView.mm, as well as accessibility implementations in other files, to be migrated at a later time.

I release the patches in this PR into the public domain.


sorbits:
Merged as dd77fcd6e...5ee17f2b0, much appreciated, thanks!

PR #1428, opened 2018-12-05
Refactor finder tags to use catalog colors

rdwampler:
Also, reworked the OFBFinderTagsChooser to have a similar look as on 10.14.


sorbits:
Looks good, merged as ca80254b80da6b4c7bccd4179fdd11ed9b46e825, thanks!

Only change I made was to have the “Tags…” label opt out of accessibility (voice over).

PR #1431, opened 2018-12-25
Implement "Symbols" accessibility rotor

dusek:
Accessibility rotors allow VoiceOver users to navigate content quickly by making possible navigation through select elements only.

In this case, we enable VoiceOver users to navigate efficiently through symbols.

The first way to invoke a rotor is a "global" one (does not take into account current position in text) and is invoked with VO-U. After pressing arrow left/right sufficient times to reach "Symbols", one is presented with a list analogical to the "Jump to symbols" TextMate feature invokable by Shift-Cmd-T. Arrow up and down then enable navigating through items, Return selects the result (and jumps to it), Esc cancels the interface without moving the cursor. Also filtering the list works by simply typing characters (implemented as case-insensitive substring search - consistent e.g. with Safari's "Headings" rotor).

The second way to use a rotor is a "local" one (navigates w.r.t. the current position in text) and is typically done through Trackpad Commander (turned on by VO + two finger "screw" gesture to the right on the Trackpad). Then one uses two-finger "screw" gesture (like (un)screwing the bottle lid from/to the bottle with thumb and index finger) to select "Symbols". When this is finally selected (the selection is persistent), the user can then use flick with one finger up or down gestures to move to the previous/next occurrence respectively.

I made a choice at one spot - I report the "symbol location - end of hardline" as the range of the symbol (that is e.g. what VoiceOver reads when navigating the elements). This seemed to work best for the user at least in the Objective-C code.

The APIs are 10.13+ only, so appropriate avaiability declarations are put in place to not cause warnings from the compiler. I was not surewhether some declaration like that is needed also for the `OakTextView`'s informal category's conformance to the `NSAccessibilityCustomRotorItemSearchDelegate` protocol - the protocol is 10.13+ only too, but the compiler does not warn about conforming to it. Perhaps the Objective-C runtime is smart enough to take into account that the `NSAccessibilityCustomRotorItemSearchDelegate` symbol is (hopefully) weakly-linked and adds conformance to it (I assume using the `class_addProtocol` function) only when it is non-nil. I have however not tested whether the compiled app actually degrades gracefully (without crashing) on macOS 10.12 and earlier, though (for starters, I seem not to have access to a macOS Sierra image to create a VM).

I release the code to the public domain.

### "Global" rotor for "Symbols"
screenshot 2018-12-25 at 17 17 49

### "Global" rotor for "Symbols" with "blink" filter applied
screenshot 2018-12-25 at 17 18 29

### After selecting item in "global" rotor for "Symbols"
screenshot 2018-12-25 at 17 19 13

### "Local" rotor for "Symbols" offered
screenshot 2018-12-25 at 17 16 07

### DIFF
--- a/Frameworks/OakTextView/src/OakTextView.mm
+++ b/Frameworks/OakTextView/src/OakTextView.mm
@@ -31,6 +31,7 @@
 #import <layout/layout.h>
 #import <ns/ns.h>
 #import <ns/spellcheck.h>
+#import <text/case.h>
 #import <text/classification.h>
 #import <text/format.h>
 #import <text/newlines.h>
@@ -279,6 +280,12 @@ struct document_view_t : ng::buffer_api_t
 		return buf.symbol_at(ranges().last().first.index);
 	}
 
+	std::map<size_t, std::string> symbols () const
+	{
+		ng::buffer_t const& buf = [_document_editor buffer];
+		return buf.symbols();
+	}
+
 	bool has_marks (std::string const& type = NULL_STR) const
 	{
 		return [_document_editor buffer].prev_mark(SIZE_T_MAX, type).second != NULL_STR;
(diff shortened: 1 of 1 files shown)


sorbits:
Merged as 85d3af94bae07982bc149440ee099970a021abb6, thanks!

For me, the first symbol in the file did not show up on VO-U so I had to [insert these lines](https://github.com/textmate/textmate/blob/85d3af94bae07982bc149440ee099970a021abb6/Frameworks/OakTextView/src/OakTextView.mm#L1815-L1818) to workaround the lack of being passed `NSNotFound`.

For the previously implemented informal accessibility methods, I believe availability macros are unnecessary because our implementation does not reference any new (10.13+) symbols, contrary to this rotor stuff that involves new API (even in the method signature).

PR #1432, opened 2019-01-10
OakSyntaxFormatter: Clean up font attributes

rdwampler:
See https://lists.macromates.com/textmate/2019-January/040883.html.

The issue is caused by commit f850dbdbc772814372ad84f8740ea8234950cc5d, which added support for font style changes. It did so by applying a single font across the entire string. But, in some cases not all characters (e.g., Chinese or Japanese characters) may be available for that font so these characters would not be rendered correctly while editing.

### DIFF
--- a/Frameworks/OakAppKit/src/OakSyntaxFormatter.mm
+++ b/Frameworks/OakAppKit/src/OakSyntaxFormatter.mm
@@ -75,6 +75,9 @@ static size_t kParseSizeLimit = 1024;
 	if(OakIsEmptyString(plain) || !_grammarName)
 		return;
 
+	if(_enabled == NO || ![self tryLoadGrammarAndTheme])
+		return;
+
 	if(!_font)
 	{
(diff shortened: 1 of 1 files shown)


sorbits:
When a string has been syntax highlighted and the regular expression option is disabled, we need to reset the styles, i.e. remove colors and use plain font (no bold or italic).

While this PR did fix the CJK issue, it would not bring the string back to a plain style, so I took just the `fixFontAttributeInRange:` bit and applied to the string (in both branches of the code, i.e. syntax highlighted and plain).

Pushed as e0ddc617945f521264380fea55452a734708d2d3.

Thanks for making me aware of `fixFontAttributeInRange:` and for your PR, sorry about the slow response.

rdwampler:
> When a string has been syntax highlighted and the regular expression option is disabled, we need to reset the styles, i.e. remove colors and use plain font (no bold or italic).

I missed that. Thanks. :)

PR #1433, opened 2019-06-15
fix name typo

studioego:
fix name typo, 'Korean – ISO-2022-JP' to 'Korean – ISO-2022-KR'.


sorbits:
Merged as 66c96b8393dcd79f1e4a773acaf8af5e8c1a963d, thanks!

PR #1434, opened 2019-06-26
Convert `.each` to `.map` and concat array after array assignment

dainmiller:
Have been a textmate user for a few years, and was just reviewing the code for performance optimizations - given I would love it to stay fast. 

Noticed this quick memory optimization in ruby. Hope it's ok!

### DIFF
--- a/bin/gen_build
+++ b/bin/gen_build
@@ -459,7 +459,6 @@ end
 
 def targets_at_path(path)
   res = [ ]
-
   new_targets = [ Target.new(path) ]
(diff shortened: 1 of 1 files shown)


sorbits:
Thanks, though I recently did a complete overhaul of that script (see 412e6798ca206b15cdc8f86a1387835a3f4f82e7).

Don’t think your patch is relevant with the new version, also, I think it should actually have been using `+=` instead of `<<` when appending an array to an the array.

PR #1441, opened 2020-02-16
Check for the “libplist” conflicting headers

fish2000:
The headers for the `libplist` open-source library† conflict with those within TextMate’s “plist” framework – specifically it has a `<plist/plist.h>` header, as well as a `<plist/Date.h>` header which will be picked up by TextMate’s build system and only generate a compiler warning instead of a hard error.

The “hard errors” come afterwards, and without explanations‡ – I have been trying to get TextMate to build for four days now without understanding why it was failing mysteriously on multitudinous files _including_ `<plist/plist.h>` while compiling `Frameworks/src/plist/plist.cc` wasn’t an issue.

This shell snippet, included in `./configure` under the “Check various dependencies” aegis, detects the `libplist` header directories, errors out if they are present, and (depending on whether or not Homebrew is available) offers a bit of terse advice about how to proceed.

Screen Shot 2020-02-16 at 9 55 03 AM

† _Figure 1:_ `libplist` _header layout._

Screen Shot 2020-02-16 at 9 55 24 AM

‡ _Figure 2: examples of errors resulting from the conflicting headers._

### DIFF
--- a/configure
+++ b/configure
@@ -82,6 +82,16 @@ for dep in ninja ragel multimarkdown pgrep pkill; do
 	which -s "$dep" || error "*** dependency missing: ‘${dep}’."
 done
 
+for libplistdir in /usr/include/plist /{opt,usr}/local/include/plist ; do
+    if [[ -d $libplistdir ]]; then
+        if which -s brew && brew ls --versions libplist > /dev/null; then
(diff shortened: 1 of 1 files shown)


sorbits:
I have addressed this by adding the path of custom headers before the more generic include directories.

Let me know if it works/fails with the open source libplist.

PR #1444, opened 2020-04-07
Miscapitalization of Menu Items

schmidhauser:
The preposition `to` in `Jump to Next/Previous Mark` should be capitalized as in `Jump to Next/Previous Bookmark`. Prepositions are generally not capitalized in title case.


sorbits:
Thanks!

PR #1447, opened 2020-07-29
Add "Close Tabs to the Left" as alternate of "Close Tabs to the Right"

belkadan:
This is useful when working with one set of files in a project, then moving to another set without closing the original set first.

Alternate menu items are pretty subtle, but there are already a lot of ways to close tabs, and I didn't want to clutter the menu further.


sorbits:
Merged as 6cd3c70b9f3b8ac236e773d3609d59d94dabee20, thanks and sorry about the late response!

PR #1448, opened 2020-12-07
Search and replace text fields changed to use NSComboBox

armartinez:
Search and replace text fields changed to use the NSComboBox class. I've removed all references to the popover and the history buttons previously used but I've kept the OakPasteboardSelector class and the helper methods in OakPasteboard, I can delete them after discussion on another commit.

### DIFF
--- a/Frameworks/Find/src/FFTextFieldViewController.h
+++ b/Frameworks/Find/src/FFTextFieldViewController.h
@@ -6,6 +6,4 @@
 @property (nonatomic) NSString* stringValue;
 
 - (instancetype)initWithPasteboard:(OakPasteboard*)pasteboard grammarName:(NSString*)grammarName;
-- (void)showHistory:(id)sender;
-- (void)showPopoverWithString:(NSString*)aString;
 @end
--- a/Frameworks/Find/src/FFTextFieldViewController.mm
+++ b/Frameworks/Find/src/FFTextFieldViewController.mm
@@ -8,7 +8,7 @@
 // = OakAutoSizingTextField =
 // ==========================
 
-@interface OakAutoSizingTextField : NSTextField
+@interface OakAutoSizingTextField : NSComboBox
 @property (nonatomic) NSSize myIntrinsicContentSize;
 @end
 
@@ -34,13 +34,12 @@
 
 static void* kFirstResponderContext = &kFirstResponderContext;
 
-@interface FFTextFieldViewController () <NSTextFieldDelegate, NSTextStorageDelegate, NSPopoverDelegate>
+@interface FFTextFieldViewController () <NSComboBoxDelegate, NSTextStorageDelegate>
 {
 	OakAutoSizingTextField* _textField;
 	OakSyntaxFormatter*     _syntaxFormatter;
 	OakPasteboard*          _pasteboard;
 	NSString*               _grammarName;
-	NSPopover*              _popover;
 }
 @end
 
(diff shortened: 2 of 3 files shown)


sorbits:
Please squash the two commits (into one) and force-push, so that we get the changed indentation noise removed.

I’ll look closer at this PR once I have wrapped up another project, I hope before Christmas.

armartinez:
Awesome! Will do 👍

sorbits:
I’m trying out this patch but on first test I notice these three issues:

1. Content is higher than frame:

    image

2. Rendering issue when string contains newlines:

    image

3. Pressing arrow down should normally open the menu, but it doesn’t do anything, except when there is a multi-line string and insertion point is at the beginning of this string.

I must confess I am skeptical myself about how feasible it is to provide a good user experience using the native `NSComboBox` control, in addition to the above issues there is also the loss of control of the menu, for example it is currently possible to delete history items using the delete key (when the menu is showing), I don’t think that can be easily supported using the native controls.

So while I really would like to switch to native controls for consistent look and feel, I am not sure it is actually worth it.

lormanian:
Hi Allan,

I commend you for your dedication to a fantastic creation. I know that
these stubborn quirks are a pain in the brain. You are a person of vision.
Happy New Year and take a break. The challenges will be there when you
are ready to get back to them.

Sincerely,

Mr. Les Ormanian
MX: +52 (614) 394-6372 (preferred)
USA: (541) 727-0333

armartinez:
@sorbits Sorry for the delay, I'm busy with other projects. I understand and share your concerns. I'm going to try to fix the issues you mentioned and see if I can get closer to the previous functionality. My idea anyways is to collaborate on the project so I just picked one of the issues described in the `Hacking the Code` wiki section. I would appreciate if you can either update that list or give me some issues that you think are relevant now to the app. Thanks!

armartinez:
@sorbits I've pushed a couple of changes to the PR. I've fixed the content being higher than the frame (the textField is not an OakAutoSizingTextField anymore), added support for using the down arrow to open the menu and also added support for using the delete key to remove history items. I'm not clear on the issue with newLines but I've noticed that pasting multiline content to the NSComboBox does present some rendering issue, but this is not exclusive to the textField, it also happens on the matching combo so it might be something worth fixing across the app.

lormanian:
With all due respect to you Axel,

I am surprised by the features and functionality of Textmate with every
use; I earnestly commend you (and your team if there is one) for the
forward-thinking and functionality inherent in Textmate. If my hiring a
"sky-writer" and the plane would boost your sales, I would; because it
really is life-altering with all of what it not only adds to a techie's
life, but what it saves in time. It may not be all slick and sexy on the
outside, but what it brings into the coding room is beyond a stunning
CPU-cycle conserving achievement

In my 40 years of being a 'techie' (think Radio Shack TRS-80 Model I Level
I and how was I ever going to use all 8K of RAM? Soldering modifications
onto the motherboard and clipping IC pins). I created network management
tools so I could stop flying from Detroit to Honduras in the middle of the
night and coded systems to detect and repair themselves. Designing and
building a power supply from scratch so our T1 and T3 lines which
terminated in the area's demark building could be on separate power backup
supplies.
Writing code to check every ethernet port on a County-wide Government 911
net, University Classrooms and Dormitories, 5000+ hybrid fiber-coax
streaming content business and residential customer environment; the
University brains who's job offer I turned down said that a program to find
out which port a machine serving apps from was impossible in our
environment. I wrote down the URL that they could visit and enter any IP or
MAC address and as a response get the exact switch and port # back. They
became almost irate insisting that it was not possible in a layer 3
switched environment. I explained that I had written and was using the code
for months; they simply needed to try the URL.


Sincerely,

Mr. Les Ormanian
MX: +52 (614) 394-6372 (preferred)
USA: (541) 727-0333

PR #1449, opened 2020-12-12
Set the frame size via NSTitlebarAccessoryViewController's view property

rdwampler:
When linking against the macOS 11.0 SDK, the view size fails to update.


sorbits:
Merged as a52383a099fdf71d43bcaacf00b73c20b00f982b, thanks!

PR #1452, opened 2021-01-16
Set Preference window toolbar style to NSWindowToolbarStylePreference…

rdwampler:
… for Big Sur

Before:
Screen Shot 2021-01-16 at 5 01 18 PM

After:
Screen Shot 2021-01-16 at 5 00 25 PM

To completely match the new preference toolbar style for Big Sur, the icons should be switched to use SF Symbols. I think we can replace all of the icons with the exception of "Bundles" with the current set of SF Symbols provided by Apple.


sorbits:
Merged as 5168278b2a2bb40bffe4615d9bfa2f1644c50c37, thanks!

I had to update `sdk-compat.h` with the proper declarations (88481bdd127873aed24bcf045310a900d165b619), it appears the define to use for the guard is `MAC_OS_X_VERSION_10_16`, at least on macOS 10.15.

PR #1462, opened 2023-02-26
SyntaxMate: Delete unused app

shpakovski:
Hi Team, this tool is not embedded into the main app and did not find a use as we expected. We should probably remove it. Thanks for understanding!


shpakovski:
If we want it to stay, could you please fork the following repos to `textmate`? Then I’ll make another pull request:

https://github.com/shpakovski/syntaxmate.tmbundle
https://github.com/shpakovski/SyntaxMateDemo

